Issue #2, opened 2019-10-25
Submit login form by pressing enter


vaaralav:

Pinqvin:
This issue was fixed in PR #4 and can be closed

PR #4, opened 2019-10-26
Add form submit handling to login view

Pinqvin:
Hooking to the form submit event allows us to utilize browser built in behaviors like submitting a form when pressing enter. This commit fixes issue #2


stevespringett:
Thank you for the PR.

PR #5, opened 2019-10-26
Fix CORS issues with devServer

vaaralav:
* Proxy API requests to the backend when using devServer.
* When developing, use relative paths for API requests. In production use absolute URIs for API requests.
* Add npm start script which starts the devServer


stevespringett:
Thanks for the PR.

Issue #1, opened 2019-10-25
First login and initial startup


vaaralav:
Add support for changing the initial admin password on first login. See the solution in existing UI: https://github.com/DependencyTrack/dependency-track/blob/c4f469bf735731a02e57e549fa9dda89abfd8977/src/main/webapp/assets/common.js#L175-L223.

> An administrative account is created on initial startup with the following credentials:
>
>username: admin
>password: admin
>Upon first login, the admin user is required to change the password.
> 
> https://docs.dependencytrack.org/getting-started/initial-startup/

Pinqvin:
I can have a look at this issue

Pinqvin:
This issue was fixed in PR #7 and can be closed

Issue #6, opened 2019-10-26
Client side form validation


Pinqvin:
Should we do more validation in the client side? Currently, there aren't that many forms in the application, but they depend on server errors for validating the input. We could do more robust validation in eg. the password force reset flow (basic checking of fields having values, but also covering password re-use, password matching etc.) before even contacting the server.

The server seems to be somewhat inconsistent on how it treats errors. For example, the `/v1/user/forceChangePassword` route sometimes returns error code strings, but can also return plain error messages. Doing validation based on the plain error messages can be a bit tricky / brittle.

I just wanted to file an issue about this to discuss if doing more robust client side validation is something we would like to do in this project and if it's okay to add eg. form validation dependencies to achieve that.

PR #7, opened 2019-10-26
Initial login password reset

Pinqvin:
Kapture 2019-10-27 at 3 54 46

This PR adds functionality for doing the initial password reset if the backend response for login dictates that we should do so. The functionality and validation in this PR is fairly rough, but should cover most basic error / success states. No form validation is done in the client side app right now, but more robust validation could be added in later. Check issue #6 for discussion on how this should be handled in the future

This PR resolves issue #1


stevespringett:
Fantastic! Thanks for the PR.

PR #10, opened 2019-10-27
Check authorization status on application initialization

### LINKED ISSUE #8: Check token validity at application init
Currently the JWT token is stored in session storage, but token validity or login status is never checked on application initialization and authorization status defaults to false. This leads to the user having to login at every page refresh.

App init should check for a saved JWT token and verify it's validity by querying the API (for example the `/v1/user/self` route or something else that every user can call). If the token exists and is valid, we should go to the requested route, otherwise we clear session storage and reroute to login

Pinqvin:
This PR adds functionality for checking the stored JWT token status during app init. Token status is checked by calling `v1/user/self` API to fetch user details. On a successful API response we set axios and jQuery authorization headers and proceed with routing etc. If an error occurs, we clear the session storage token and redirect to the login page.

Also changes successful login attempts to set the axios and jQuery authorization headers.

Fixes #8

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -6,28 +6,48 @@
   import axios from 'axios';
   // bootstrap-table still relies on jQuery for ajax calls, even though there's a supported Vue wrapper for it.
   import $ from 'jquery';
+  import api from './shared/api';
   export default {
     name: 'app',
     data() {
       return {
         authenticated: false,
+        user: null,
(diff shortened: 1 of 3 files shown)


stevespringett:
Thanks for the PR

stevespringett:
Thanks for the PR

PR #11, opened 2019-10-28
Add client side validation to the login form

### LINKED ISSUE #6: Client side form validation

Pinqvin:
loginvalidation

This PR showcases how we could do client side validation for the forms by doing some required validations in the login form. Validation is done with the VeeValidate library.

Admittedly, the validation adds quite a bit of noise to the actual login view, but validation could also be encapsulated into separate input components. This way, every form would just require the validation observer wrapping to make sure that invalid forms aren't submitted, but the actual validation provider wrapping and Bootstrap related state checking would be done inside the individual input components. Since the Login form is so small, I didn't do the component separation yet.

Fixes #6


stevespringett:
Thank you for the PR.

PR #12, opened 2019-11-03
Refactor validation to a separate component

Pinqvin:
This PR refactors validation to happen in separate form input components. It also adds more thorough validation to the password force change flow


stevespringett:
Nice. Thanks for the PR.

PR #17, opened 2020-03-16
Updates core-js to version 3.

barslev:
A review of the code indicates that no other parts are affected by the breaking changes in the library update.


fnxpt:
doesnt make sense anymore... this is already on a higher version

PR #18, opened 2020-03-22
Add OpenID Connect Support

nscuro:
This PR integrates the OpenID Connect capabilities that have been implemented in Dependency-Track with https://github.com/DependencyTrack/dependency-track/pull/626. A way to test this with Dependency-Track and Keycloak has been documented in https://github.com/stevespringett/Alpine/pull/78.

#### Login view

The login view now includes an additional button for logging in with OpenID Connect. This button is only visible when OpenID Connect is enabled and configured in both backend and frontend.

Login view

Clicking this button will cause a redirect to the configured identity provider. Here: a local keycloak instance.

Identity provider login

#### Administration

The administration section now includes the pages *OpenID Connect Users*

Admin: OpenID Connect users

 and *OpenID Connect Groups*

Admin: OpenID Connect groups

Mappings of OIDC groups to Dependency-Track teams can be edited either in the *OpenID Connect Groups* page as shown above or in the *Teams* view.

Admin: Teams

#### Miscellaneous

Instantiation of `Vue` was moved. It now happens only **after** `config.json` has been loaded. Before, I was experiencing race conditions where `oidcUserManager` in `Login.vue` was instantiated before `config.json` was loaded and processed. See https://forum.vuejs.org/t/runtime-config-loading/38059/6

### OpenID Connect implementation

We're using [`oidc-client`](https://github.com/IdentityModel/oidc-client-js) to do the heavy lifting for us. This allows us to support two major OpenID Connect flows: [*Code with PKCE*](https://oauth.net/2/pkce/) and [*Implicit*](https://oauth.net/2/grant-types/implicit/). Even though *Implicit* is considered to be deprecated by *Code + PKCE*, support was added because PKCE is only supported in Keycloak since version 7. Commercial Keycloak distributions like [Red Hat SSO are currently based on Keycloak 4](https://access.redhat.com/articles/2342881).

The flow can be configured via the configuration value `OIDC_FLOW`. If this doesn't explicitly say `implicit` (lol), *Code + PKCE* is assumed.


nscuro:
A question upfront @stevespringett: How do we actually make `config.json` customizable by users? Especially in case of deployments with Docker.

stevespringett:
The UI can be deployed independently of the server. So the UI could be deployed to an NGINX or Apache webserver. The config.json would be modified in this scenario which would point to the DT server.

stevespringett:
There isn't a way currently to modify the config.json in the container itself. The config.json is deployed inside the root of the executable jar itself. This may be an opportunity to enhance the executable war library and add a path to the webapp and then make that an environment setting in the Docker container.

PR #19, opened 2020-04-08
Use the group word for notification group

roycyt:
Before:
alert_before
After:
alert_after


stevespringett:
nice catch. thanks for the pr.

PR #27, opened 2020-06-26
Reload audit trail after analysis modifications

nscuro:
This PR addresses an issue where the audit trail would not reload after performing analysis actions. Because the analysis data is returned for PUT operations as well, there's no additional overhead.

Before:

before

After:

after


stevespringett:
Ported changes to ProjectFindings over to new ProjectPolicyViolations. Did not include ComponentVulnerability since that functionality was removed from frontend 1.1

PR #33, opened 2020-08-15
Add Docker support

nscuro:
This PR adds rudimentary docker support.

Runtime configuration is made possible by either specifying environment variables or
by mounting a `config.json` from the host system. Environment variables are applied to
the container's `config.json` on startup, **but only if** `config.json` was not mounted as volume.

I also extended the GitHub workflow to build and publish the Docker image.
Publishing should only be done for pushes to master and only for the newest (`12.x`) version of Node.

`nginx.conf` may need some tweaking, it is currently pretty much identical to the one in [Vue's documentation](https://cli.vuejs.org/guide/deployment.html#docker-nginx). I didn't notice anything wrong until now though, so it might just as well be totally fine.


nscuro:
Note that the provided `docker-compose.yml` is mainly for testing. It'd probably make more sense to provide a polished version of it in Dependency-Track's main repo once this image has been pushed to Docker Hub.

stevespringett:
Again, thanks for the quality PR. I think some of this, as well as a few other things, may need to be documented. BTW, I've modified the GitHub pages for docs.dependencytrack.org so that it doesn't publish on commit to master, only the 3.8 branch.

nscuro:
Thanks @stevespringett. I already started to update the OIDC documentation to include instructions for the frontend. Might be a good opportunity to update the Docker deployment page as well.

PR #42, opened 2020-10-27
Make oidc scopes configurable

nscuro:
Some providers require OIDC clients to request more than the default scopes. Auth0 for example has extra scopes for `groups`, `roles` and `permissions`: https://auth0.com/docs/extensions/authorization-extension/use-rules-with-the-authorization-extension


stevespringett:
Thanks @nscuro. No server changes correct?

nscuro:
@stevespringett No, the scopes are transparent for the server. They just control what kind of information the server can access when calling the `/userinfo` endpoint with the token requested by the frontend.

Issue #41, opened 2020-10-19
Link to project page by name and version


sephiroth-j:
Project pages should also be reachable by providing the name and version. 

### Current Behavior:
The project page has a URL of type `/projects/{uuid}`. [SVG Badges](https://docs.dependencytrack.org/integrations/badges/) can be used with name and version using the URL pattern `/api/v1/badge/vulns/project/{name}/{version}`. In this case, it is not possible to provide a link to the project page because this would require the knowledge of the uuid. If the uuid would be known, the SVG badge would have been used with the uuid.

### Proposed Behavior:
It should also be possible to reach a project page using a URL of type `/projects/{name}/{version}`.

stevespringett:
PR's are welcome

sephiroth-j:
Will do. Is there a scheduled release date?

stevespringett:
4.0 (includes frontend v1.1) is feature complete and is being tested and upgrade logic written. The release of 4.0 is likely still a month away. This would need to be added after 4.0 is released.

kkretsch:
For maintaining a gitlab badge it might also be nice to have a "latest" version as an option.
I'd like to link to the must up2date svg badge from the project status page, without knowing the exact version and updating that link each time.

How is the change progress anyhow?

sephiroth-j:
> For maintaining a gitlab badge it might also be nice to have a "latest" version as an option. I'd like to link to the must up2date svg badge from the project status page, without knowing the exact version and updating that link each time.
> 
I think this is the same as DependencyTrack/dependency-track/issues/587

At the moment it is not possible to add such an URL [because the version can be an arbitrary string that does not need to follow a versioning schema](https://github.com/DependencyTrack/dependency-track/issues/378#issuecomment-505909856). Without such a schema, the latest could be "that version with the newest SBOM upload". But this might not be what others would expect. And when you upload a SBOM for version 1.0.0 after the one for version 2.0.0, you get the results of version 1.0.0.

> How is the change progress anyhow?

PR #45 needs a rebase. But before that, I would like to have a decision on whether my approach is going in the right direction or not. Personally, I think that using the router and named routes is more robust/better approach than hard-coding URLs and just adding a redirect from `/projects/:name/:version`  to ` /projects/:uuid`.

PR #45, opened 2020-11-25
Link to project page by name and version

sephiroth-j:
support URLs of type `projects/:name/:version`
#41

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -63,6 +63,16 @@ function configRoutes() {
           }
         },
         {
+          path: 'projects/:name/:version',
+          name: 'Project',
+          props: (route) => ( { name: route.params.name, version: route.params.version } ),
+          component: Project,
+          meta: {
+            i18n: 'message.projects',
+            sectionPath: '/projects'
+          }
+        },
+        {
           path: 'components',
           name: 'Component Lookup',
           component: ComponentSearch,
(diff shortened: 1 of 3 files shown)


stevespringett:
Thanks for the PR. 

All links including clicking on a component or vulnerability no longer work due to the the additional path in the URL.

Examples:
http://localhost:8081/projects/components/8d037fa4-57cc-4fe9-a5ef-d21c6fb8c90a
http://localhost:8081/projects/vulnerabilities/NVD/CVE-2010-5321

Both of these are invalid. I would recommend changing the logic when performing the name/version lookup that upon a successful lookup, to replace the /projects/name/version with /projects/uuid. That way, all existing logic to construct links continues to work.

sephiroth-j:
Well, the existing logic to construct links was the actual problem. I missed them because I was searching for references to the router but did not found them. Now the router is used to resolve all links by using the name of the route.

msymons:
@sephiroth-j,  is this PR still something that you want to see merged?

If so, when resolving conflicts, remember that commits must now be signed.  I think that DCO checks on PRs was enabled since you submitted this PR.

sephiroth-j:
@msymons , thanks for the reminder. I would like to know if named routes should not be used at all and the redirect approach should be used. The router with it's named routes has, for example, the benefit that query and path arguments gets url-encoded by default. Named routes can be maintained at a central place and there is not struggle with (wrong) relative paths when using the router with named routes. Depending on this decision, I would discard this PR and create a new one that just includes the redirect.

msymons:
One for @nscuro or @stevespringett to answer!  :smile:

Issue #20, opened 2020-04-28
Login redirect doesn't work


llamahunter:
The defect may already be reported! Please search for the defect before creating one.

### Current Behavior:

When deep linking into a page in deptrack, you first get redirected to the login page (despite being logged in already), and then you get deposited into the overview page, rather than the deep link you were following.

### Steps to Reproduce:

- Log in to deptrack.
- Copy a deep link to some project.
- Paste link into a new browser tab

### Expected Behavior:

Go directly to the project page.

### Environment:
  
  - Dependency-Track Version: v3.8.0
  - Distribution: Docker
  - BOM Format & Version: SBOM?
  - Database Server: PostgreSQL
  - Browser: Chrome

### Additional Details:

Login is configured to use LDAP

stevespringett:
Looks like the initial load fails and redirects to login screen, but repasting the URL into the same tab works.

llamahunter:
Also, the login page should send you back to the deep link after login is successful, not the Overview page.

sephiroth-j:
@stevespringett: You can close this one as well after #47 has been merged.

qhantom:
@stevespringett Can be closed can't it?

PR #47, opened 2020-12-06
redirect to original location upon successful login

sephiroth-j:
- stores the location before the user is redirected to the login page in the query param "redirect". after successful login, the user gets redirected to this url if it is set. navigation flow will be as follows:
1. http://localhost:8081/projects/285384c9-a18f-465f-921b-d465efd9c43e
2. http://localhost:8081/login?redirect=%2Fprojects%2F285384c9-a18f-465f-921b-d465efd9c43e
3. http://localhost:8081/projects/285384c9-a18f-465f-921b-d465efd9c43e

- uses global navigation guard to ensure the jwt is valid and redirects to login if it is not.
previous implementation checked the jwt only when the app was created. if it was expired afterwards, the app did not recognized that. the guard will take care of this now.

- some other small fixes:
  - "go back" on 404-page always directed the user to "/" --> goes back now as the text implies
  - modal dialog to edit user profile starts fetching the user from the back-end before it already visible. minimizes the effect that the input fields are initially empty and will be filled with content after the dialog is already visible.
  - wrong types for data values in `ProfileEditModal`

references #20

### DIFF
--- a/src/views/pages/Login.vue
+++ b/src/views/pages/Login.vue
@@ -75,6 +75,7 @@ import { ValidationObserver } from "vee-validate";
 import BValidatedInputGroupFormInput from "../../forms/BValidatedInputGroupFormInput";
 import InformationalModal from "../modals/InformationalModal";
 import EventBus from '../../shared/eventbus';
+import { getRedirectUrl } from '../../shared/utils';
 const qs = require("querystring");
 
 export default {
@@ -86,8 +87,8 @@ export default {
   },
   data() {
     let redirectUri = `${ window.location.origin }/static/oidc-callback.html`;
-    // redirect to url from query param but only if it starts with / so we do not redirect to external sites
-    const redirectTo = this.$router.currentRoute.query.redirect && this.$router.currentRoute.query.redirect.startsWith('/') ? this.$router.currentRoute.query.redirect : undefined;
+    // redirect to url from query param but only if it is save for redirection
+    const redirectTo = getRedirectUrl(this.$router);
     if (redirectTo) {
       redirectUri += `?redirect=${ encodeURIComponent(redirectTo) }`;
     }
@@ -122,8 +123,8 @@ export default {
           "Content-Type": "application/x-www-form-urlencoded"
         }
       };
-      // redirect to url from query param but only if it starts with / so we do not redirect to external sites
-      const redirectTo = this.$router.currentRoute.query.redirect && this.$router.currentRoute.query.redirect.startsWith('/') ? this.$router.currentRoute.query.redirect : undefined;
+      // redirect to url from query param but only if it is save for redirection
+      const redirectTo = getRedirectUrl(this.$router);
       axios
         .post(url, qs.stringify(requestBody), config)
         .then(result => {
@@ -204,8 +205,8 @@ export default {
             .then(result => {
               if (result.status === 200) {
                 EventBus.$emit('authenticated', result.data);
-                // redirect to url from query param but only if it starts with / so we do not redirect to external sites
-                const redirectTo = this.$router.currentRoute.query.redirect && this.$router.currentRoute.query.redirect.startsWith('/') ? this.$router.currentRoute.query.redirect : undefined;
+                // redirect to url from query param but only if it is save for redirection
+                const redirectTo = getRedirectUrl(this.$router);
                 redirectTo ? this.$router.replace(redirectTo) : this.$router.replace({ name: "Dashboard" });
               }
             })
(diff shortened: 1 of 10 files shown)


stevespringett:
@nscuro what impact will this PR have on OIDC flow?

nscuro:
@stevespringett Shouldn't have an impact on OIDC as far as I can tell from just looking at it, other than it simply not working with OIDC in its current state. `oidc-callback.html` would need to be updated to handle the new `redirect` param.

nscuro:
Changing the password (e.g. upon first login with `admin:admin`) doesn't work. The error is: 

```
Uncaught (in promise) Error: Redirected when going from "/login?redirect=%2Fdashboard" to "/change-password" via a navigation guard.
    zt vue-router.esm.js:1958
    Lt vue-router.esm.js:1917
    p vue-router.esm.js:2266
    56d7 index.js:249
    p vue-router.esm.js:2252
    r vue-router.esm.js:1898
    r vue-router.esm.js:1902
    Ft vue-router.esm.js:1906
    confirmTransition vue-router.esm.js:2282
    transitionTo vue-router.esm.js:2152
    replace vue-router.esm.js:2531
    replace vue-router.esm.js:2963
    replace vue-router.esm.js:2962
    login Login.vue:131
    promise callback*login Login.vue:128
    handleSubmit vee-validate.esm.js:1875
    l vee-validate.esm.js:66
    s vee-validate.esm.js:47
    a vee-validate.esm.js:37
    promise callback*l vee-validate.esm.js:39
    o vee-validate.esm.js:40
    o vee-validate.esm.js:36
    handleSubmit vee-validate.esm.js:1865
    submit Login.vue:3
```

sephiroth-j:
@nscuro , thanks for looking at this. I have fixed the error with the "PasswordForceChange" route. The redirection will also work now with this route.

I will have a closer look at the OIDC flow.

sephiroth-j:
> Would it be a viable option to store the redirect target in `localStorage`? As far as I understand it, `redirect` is intendet to be always set by the application and never by the user.

If you open multiple tabs while being logged out and if it stored in `localStorage` or `sessionStorage`, the last opened tab overrides the values for the other tabs. It would also be necessary to remove the value from the storage when it is not needed anymore. Therefore it is better to set the value only then and there, where it is needed.

nscuro:
@sephiroth-j Makes sense, thanks for clarifying.

sephiroth-j:
I have made a few additions to the OIDC flow. @nscuro , would you please test it?

src/views/pages/Login.vue
```diff
@@ -85,6 +85,13 @@ export default {
     ValidationObserver
   },
   data() {
+    let redirectUri = `${ window.location.origin } /static/oidc-callback.html`;
```
nscuro, line 88: This produces an invalid URL due to the space between `}` and `/static`.

nscuro:
@sephiroth-j Works great, besides the typo in `redirectUri` construction. I'll have to update the OIDC docs though, because IdPs like Keycloak do strict matching against the `redirect_uri` parameter. If `http://dependencytrack.example.com/static/oidc-callback.html` is configured, and `http://dependencytrack.example.com/static/oidc-callback.html?redirect=/vulnerabilities` is provided, the IdP will refuse to authenticate.

src/views/pages/Login.vue
sephiroth-j: fixed it

sephiroth-j:
> I'll have to update the OIDC docs though, because IdPs like Keycloak do strict matching against the `redirect_uri` parameter. If `http://dependencytrack.example.com/static/oidc-callback.html` is configured, and `http://dependencytrack.example.com/static/oidc-callback.html?redirect=/vulnerabilities` is provided, the IdP will refuse to authenticate.
Hm, maybe `sessionStorage` is an option here after all.

nscuro:
Shouldn't be an issue. That was supposed to be more like a TODO for me. In Keycloak you can use wildcards for stuff like this, and I'd expect that other IdPs offer similar functionality as well.

sephiroth-j:
I made the checks more restrictive before using the redirect url. Unfortunately, they still don't pass the CodeQL check - I would consider this a false positive.

nscuro:
Looks good to me. Could you include this in 1.3.0 @stevespringett?

stevespringett:
Thank you @sephiroth-j and @nscuro. Really good work here.

37b:
@nscuro 
> Shouldn't be an issue. That was supposed to be more like a TODO for me. In Keycloak you can use wildcards for stuff like this, and I'd expect that other IdPs offer similar functionality as well.

This is an issue for Azure AD and supposedly the OAuth2.0 spec [rfc6749#section-3.1.2](https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2) though I think Microsoft's reading of it may be stricter than intended - not sure as this is not my domain. 

Either way, OIDC is only partially working in DTrack with Azure AD as it does not allow wildcards in return URIs. Can the original location be stored elsewhere?

PR #56, opened 2021-02-13
Add meaningful page titles

### LINKED ISSUE #55: Pages should have a meaningful title
### Current Behavior:
Each page has "Dependency-Track" as its title. Navigating through the browsing history when every page has the same title is terrible.

### Proposed Behavior:
Unique pages should have their own page title.

sephiroth-j:
This adds page titles of type "Dependency-Track - xxx" for static pages and dynamic content such as projects, vulnerabilities, components and licenses.

fixes #55

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12567,6 +12567,11 @@
         "vue-style-loader": "^4.1.0"
       }
     },
+    "vue-page-title": {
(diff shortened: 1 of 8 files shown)


stevespringett:
Thanks for the PR

PR #61, opened 2021-03-02
add a job to excersise compose

davidkarlsen:


fnxpt:
Not really sure if running docker-compose is a valid test

PR #105, opened 2021-12-21
Make http port modifiable

florentulve:
Deptrack frontend and server both listened 8080 port by default. 
Deploying frontend and server within a podman pod is not possible without modify some namespace sharing configuration as mentionned here [https://stackoverflow.com/questions/60558185/is-it-possible-to-run-two-containers-in-same-port-in-same-pod-by-podman](https://stackoverflow.com/questions/60558185/is-it-possible-to-run-two-containers-in-same-port-in-same-pod-by-podman)

Thus it would be great to allow changing port.

This proposal use nginx env sub to handle this (template+envsubst).
envsubst doesn't handle default value, this is handle by a custom entrypoint as describe here [https://serverfault.com/questions/577370/how-can-i-use-environment-variables-in-nginx-conf](https://serverfault.com/questions/577370/how-can-i-use-environment-variables-in-nginx-conf)


fnxpt:
not really sure if it makes sense, with docker you can map 8080 to a different port, so internally it will be 8080 but to the host it can be whatever you want

PR #110, opened 2022-01-22
Disable upload BOM button when no file selected

### LINKED ISSUE #86: Deactivate Upload Button Until File Chosen
### Current Behavior:

Dependency-Track provides the ability to "Upload BOM" via a Projects's "Components" tab.

If one clicks the "Upload" button on the dialog without first selecting a file then DT displays a "BOM Uploaded" message and closes the dialog.... even though nothing was actually uploaded  (tested using v4.2.2).

image

The api-server log reports:

```
WARN [BomUploadProcessingTask] The BOM uploaded is not in a supported format.
 Supported formats include CycloneDX XML and JSON
```

### Proposed Behavior:
The "Upload" button on the dialog should be inactive until a file is selected.

sekwah41:
Fixes #86

image

image


stevespringett:
Thanks for the PR.

Issue #94, opened 2021-09-01
Add Policy Violations Column to Projects Page


msymons:
### Current Behavior:
The Projects page offers the following columns:

* Project Name
* Version
* Last BOM Import
* BOM Format
* Risk Score
* Active
* Vulnerabilities

ie, no display of policy violations

### Proposed Behavior:
Add a sortable "Policy Violations" column to the page.  This will make it easy to perform tasks such as:
* Sort by risk and check for low policy violaton count...  which may be an indication that policies are incorrectly configured.
* Sort by policy violation count to see the worst offending projects.  When policies are "mature" then these projects should perhaps be focused on first.  

Note that a simple count of total violations would include License and Operability Risk...  so maybe a bit more nuance might be needed.

sekwah41:
While all the data is there to visually add them, to sort them I think would require a backend change to store the policy violations total to store a value like `lastPolicyViolationsTotal`.

I suggest copying the visuals from the vulnerabilities tab and then using a [tooltip](https://bootstrap-vue.org/docs/components/tooltip) to allow users to hover over to get more information about it.

```javascript
{
  policyViolationsAudited: 0,
  policyViolationsFail: 0,
  policyViolationsInfo: 0,
  policyViolationsLicenseAudited: 0,
  policyViolationsLicenseTotal: 0,
  policyViolationsLicenseUnaudited: 0,
  policyViolationsOperationalAudited: 0,
  policyViolationsOperationalTotal: 0,
  policyViolationsOperationalUnaudited: 0,
  policyViolationsSecurityAudited: 0,
  policyViolationsSecurityTotal: 0,
  policyViolationsSecurityUnaudited: 0,
  policyViolationsTotal: 0,
  policyViolationsUnaudited: 0,
  policyViolationsWarn: 0
}
```

Ill make a preview using ![#20a8d8](https://via.placeholder.com/15/20a8d8/000000?text=+), Warn ![#ffc107](https://via.placeholder.com/15/ffc107/000000?text=+) and Fail ![#f86c6b](https://via.placeholder.com/15/f86c6b/000000?text=+)

As having too many different colours may get a bit confusing to be useful information.

I was thinking the violation types may be good originally but then wed need colors to represent violation types which I don't believe there are.

sekwah41:
Proposed view

![image](https://user-images.githubusercontent.com/3209834/150649619-44e5ef94-b530-4324-95fe-4224ab845d6f.png)
Just to also make it a little clearer

[...]

PR #111, opened 2022-01-22
Policy violations

sekwah41:
Also rewritten how the progress bar was done to better use bootstrap-vue rather than calculating values in js.

Before (Progress bar changes)
image

After (Progress bar changes)
image

Added new column and hover info to make it clearer
image
Just to also make it a little clearer

I originally wanted to make it go above the bar though that messes up the view for the top entry as it detects the edge of the table as a boundary.


src/views/components/PolicyViolationProgressBar.vue
```diff
@@ -0,0 +1,39 @@
...
+  props: {
+    total: Number,
+    info: Number,
+    warn: Number,
+    fail: Number,
+  },
+  data() {
+    return {
+      // Workaround for vue references to the progress-bars. Using the ref targets doesn't seem to work.
+      hoverId: Math.random().toString(36),
```
sekwah41, line 35: See https://bootstrap-vue.org/docs/components/tooltip

For some reason when trying to use the ref it seems to get a reference but is referencing old versions of the components no longer used. Hence the hover fields don't work. Though the direct reference method works so I am generating a reasonably unique id that doesn't have to be secure or anything to avoid the tooltips showing on the wrong components.

sekwah41:
Should partially implement #94

Though for filtering based on total count I believe a backend change is needed.

sekwah41:
Need to add translation strings for the messages. Though can personally flag the pr as not ready now.

sekwah41:
Updated with correct translation strings 
image

sekwah41:
Could also possibly add more info such as the violation category totals. Just for a little more hover info. Rather than just reshowing what is already there.

sekwah41:
Could also show more tooltip info like this. Though not merged atm 
image

stevespringett:
Thanks for the quality PR. Is this ready to test and merge?

sekwah41:
@stevespringett should be unless you want any visual changes. I've also added translation strings for each.

PR #112, opened 2022-01-22
Reduce hit radius of graphs to avoid #92

### LINKED ISSUE #92: Cannot display Tooltip info for latest point on graphs
### Current Behavior:

When hovering the mouse pointer over any graph in Dependency-Track (the main dashboard has the most graphs) then one can point at "now" (the point on the far right of any graph) and yet the tooltip data only shows info from a few days ago.

image

image

### Steps to Reproduce:
Wave your mouse pointer over any DT graph.  When pointing at the right-most point on any graph,  the displayed tooltip data is not current,  It represents a point on the graph to the left of where your mouse is pointing.

Using Zoom (in or out) does not help.

### Expected Behavior:
Mouse-over should support displaying data relevent to whatever point on on the graph the mouse is hovering over,

### Environment:
  
  - Dependency-Track Version: 3.8.0, 4.3.3
  - Client Browser:   Firefox, Chrome, Microsoft Edge
  - Client O/S: Windows 10, Mac.

sekwah41:
Should fix #92 or at least make it more usable. It may be a little awkward for users used to the old large hover hitboxes though will avoid the issues as seen in #92.

### DIFF
--- a/src/views/dashboard/ChartAuditingProgress.vue
+++ b/src/views/dashboard/ChartAuditingProgress.vue
@@ -88,7 +88,7 @@
             },
             point: {
               radius: 0,
-              hitRadius: 100,
+              hitRadius: 20,
               hoverRadius: 4,
               hoverBorderWidth: 3
             }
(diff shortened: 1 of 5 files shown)


stevespringett:
Thanks for the PR

PR #116, opened 2022-02-05
Add `VIEW_VULNERABILITY` permission

nscuro:
Companion PR to https://github.com/DependencyTrack/dependency-track/pull/1373.

* Makes the audit tab in the project view visible to users with `VIEW_VULNERABILITY` permission.
* Hides input fields in the audit tab for users without `VULNERABILITY_ANALYSIS` permission


stevespringett:
Thanks for the PR

PR #118, opened 2022-02-10
Switch to sass from node-sass to widen node version support

sekwah41:
This is in response to #117 which is not supported in the current automated builds.

Doesn't seem to cause any issues and in the past, I've swapped it out and it's been fine.

Though there are a few deprecation warnings it does state in all of the files what needs to be switched and they are all within font awesome not using updated SCSS generally. The new one is documented here https://sass-lang.com/documentation/breaking-changes/slash-div though they could probably use just raw css [calc](https://caniuse.com/calc).

Though if it gets dropped before they update it would be a simple pr to font-awesome to fix.


sekwah41:
Might want to look into this @stevespringett but there isn't a test script in the package.json. Just went back to look at past pr's and it just doesn't exist and doesn't fail despite the error message. Its only in node 16 that they seem to return that the not found exists with a proper error code.

PR #121, opened 2022-02-15
Disable complementary analysis fields until a state is set

nscuro:
The API server only really records anything besides comments and the analysis state *after* the state has been changed. 

It also makes sense that the complementary fields are only set after the analysis state changed, and the frontend should reflect this intended workflow.

https://user-images.githubusercontent.com/5693141/154135189-2c853ced-1627-4147-9872-450e667e6cfe.mp4


stevespringett:
Nice! Thanks for the PR.

PR #141, opened 2022-04-25
ci: overhaul and release automation

### LINKED ISSUE #78: Migrate Docker image from Alpine to Debian Slim
Alpine images can cause some incompatibility issues due to its use on musl instead of glibc.

This ticket is to track the migration of the Dockerfile from using an Alpine base image to one based on Debian slim.

Related to https://github.com/DependencyTrack/dependency-track/issues/1054

API Server: https://github.com/DependencyTrack/dependency-track/issues/1090

k3rnelpan1c-dev:
### Description

In the same veins as DependencyTrack/dependency-track#1419, this PR offers a CI overhaul and migrates the current release workflow from a bash script based solution to a GHA Workflow.

I took the liberty to include a proposal configuration for [renovate-bot](https://docs.renovatebot.com/), as I personally find that a better configurable dependabot alternative for frontend repositories. [An example of what this config would produce](https://github.com/k3rnelpan1c-dev/frontend/pulls)

### Changes

* update current CI workflows
* overhaul the current Containerfile
* add a Debian flavoured Containerfile
* add a release workflow (can handle pre- and regular rerelease)
* ~~propose a [renovate-bot](https://docs.renovatebot.com/) config as it supports grouping and very granular configuration~~ (opted to move to followup PR)

### Issues

* ~~addresses #78 (by adding a Debian flavour)~~
* closes #78 by discussing and opting to keep Alpine Linux for the frontend


.github/renovate.json
```diff
@@ -0,0 +1,60 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:base",
+    ":prHourlyLimit4",
+    ":prConcurrentLimit20",
+    ":pinAllExceptPeerDependencies",
+    ":semanticCommits",
+    ":semanticCommitScope(deps)",
```
nscuro, line 9: Are [semantic commits](https://www.conventionalcommits.org/en/v1.0.0/) something you want to have, @stevespringett?  If they are used, they should be used consistently, otherwise they're kinda pointless. This would also involve stuff like [commit squashing](https://www.conventionalcommits.org/en/v1.0.0/#do-all-my-contributors-need-to-use-the-conventional-commits-specification) when merging PRs etc.

If we don't commit (lol) to using it, it shouldn't be included in the renovate config.

.github/workflows/ci-release.yaml
```diff
@@ -0,0 +1,112 @@
...
+        uses: actions/setup-node@v3.1.1
+        with:
+          node-version: '16'
+          cache: 'npm'
+
+      - name: Setup Environment
+        id: variables
+        run: |-
+          git config user.name "dependency-track-bot"
+          git config user.email "dependency-track-bot@owasp.com"
```
nscuro, line 38: Just curious: Why is this necessary? How does this behave compared to [what's done in the Dependency-Track pipeline](https://github.com/DependencyTrack/dependency-track/pull/1419/files#diff-eecbed2b66fe9be749f388e527e472f10b086c8cced6e90c234c8a23be5b8798R88-R93) via the `gh` CLI? Do we have to create that user first?

docker/Dockerfile.alpine
```diff
@@ -0,0 +1,47 @@
...
+COPY ./docker/etc/nginx/conf.d/default.conf /etc/nginx/conf.d/default.conf
+COPY ./docker/docker-entrypoint.sh /docker-entrypoint.d/30-oidc-configuration.sh
+
+# Specify the container working directory
+WORKDIR ${APP_DIR}
+
+# metadata labels
+LABEL \
+    org.opencontainers.image.vendor="OWASP" \
+    org.opencontainers.image.title="Official Dependency-Track Container image" \
```
nscuro, line 40:
```suggestion
    org.opencontainers.image.title="Official Dependency-Track Frontend Container image" \
```

docker/Dockerfile.alpine
```diff
@@ -0,0 +1,47 @@
+FROM nginxinc/nginx-unprivileged:1.20.2-alpine
```
nscuro, line 1: Can we pin the digest like we're doing it for the API server? https://github.com/DependencyTrack/dependency-track/blob/master/src/main/docker/Dockerfile#L1

docker/Dockerfile.debian
```diff
@@ -0,0 +1,47 @@
...
+COPY ./docker/etc/nginx/conf.d/default.conf /etc/nginx/conf.d/default.conf
+COPY ./docker/docker-entrypoint.sh /docker-entrypoint.d/30-oidc-configuration.sh
+
+# Specify the container working directory
+WORKDIR ${APP_DIR}
+
+# metadata labels
+LABEL \
+    org.opencontainers.image.vendor="OWASP" \
+    org.opencontainers.image.title="Official Dependency-Track Container image" \
```
nscuro, line 40:
```suggestion
    org.opencontainers.image.title="Official Dependency-Track Frontend Container image" \
```

docker/Dockerfile.debian
```diff
@@ -0,0 +1,47 @@
+FROM nginxinc/nginx-unprivileged:1.20.2
```
nscuro, line 1: Can we pin the digest like we're doing it for the API server? https://github.com/DependencyTrack/dependency-track/blob/master/src/main/docker/Dockerfile#L1

docker/docker-entrypoint.sh
```diff
@@ -4,25 +4,15 @@ set -e
...
-
-    # Override default config file
-    mv -f /tmp/config.json ./static/config.json
+  sed -i ./static/config.json \
+    -e "s;\"API_BASE_URL\": \"\";\"API_BASE_URL\": \"${API_BASE_URL}\";" \
+    -e "s;\"OIDC_ISSUER\": \"\";\"OIDC_ISSUER\": \"${OIDC_ISSUER}\";" \
+    -e "s;\"OIDC_CLIENT_ID\": \"\";\"OIDC_CLIENT_ID\": \"${OIDC_CLIENT_ID}\";" \
+    -e "s;\"OIDC_SCOPE\": \"openid email profile\";\"OIDC_SCOPE\": \"${OIDC_SCOPE}\";" \
+    -e "s;\"OIDC_FLOW\": \"code\";\"OIDC_FLOW\": \"${OIDC_FLOW}\";" \
+    -e "s;\"OIDC_LOGIN_BUTTON_TEXT\" : \"\";\"OIDC_LOGIN_BUTTON_TEXT\" : \"${OIDC_LOGIN_BUTTON_TEXT}\";"
```
nscuro, line 15: I feel like this is a step back. Not only is it harder to read, it may also break when the formatting on `config.json` changes, or the current default values change.

.github/renovate.json
```diff
@@ -0,0 +1,60 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:base",
+    ":prHourlyLimit4",
+    ":prConcurrentLimit20",
+    ":pinAllExceptPeerDependencies",
```
nscuro, line 7: What's the reasoning behind this? It doesn't look like we currently have any `peerDependencies`.

.github/renovate.json
```diff
@@ -0,0 +1,60 @@
...
+    ":semanticCommits",
+    ":semanticCommitScope(deps)",
+    "docker:disableMajor",
```
nscuro, line 10: I think we want to get notified about major updates as well. Especially with NGINX, there's typically not much that can go wrong.

nscuro (changes requested):
Overall this looks solid, just a few minor comments that may need clarification.

Also huge +1 for using Renovate, it's *much* better than Dependabot. It will need to be activated first however.

.github/renovate.json
k3rnelpan1c-dev: I second that! As this was just an initial proposal to see if this was something you would be interested I reused some of my other configs I use personally, which included semantic commits :sweat_smile: 

Depending if you want to commit to it (pun intended) or not I would keep or remove this preset.

.github/workflows/ci-release.yaml
k3rnelpan1c-dev: well, this was necessary to quiet `npm version`, which implicitly commits the version bump and would complain about an unset committer otherwise.

However, this can be worked around by telling npm not to implicitly commit, and committing via `gh` similar to the backend CI.

docker/Dockerfile.alpine
k3rnelpan1c-dev: ofc! just dint bother yet as Renovate would have pined it for me xD (bad excuse for me being to lacy to grab the digest hashes, but will do now)

docker/docker-entrypoint.sh
k3rnelpan1c-dev: while I see your point, I ended up changing it to this as it allowed me to skip the only package manager call of the image. Additionally, the fact that it could be easily realised via a `sed` call and therefor be done with "on board" tools nudged me to alter it.

Anyway, I don't feel too strongly about it and its an easy revert :man_shrugging:

.github/renovate.json
k3rnelpan1c-dev: ah, yes, while you don't use any `peerDependencies` you do utilise the `engines` property.

see: https://docs.renovatebot.com/presets-default/#pinallexceptpeerdependencies

if you prefer a specific `packageRule` over using this preset to skip pinning the `engines` property I can change it to that.

.github/renovate.json
k3rnelpan1c-dev: fair, this on is another of my "copy-pasta" config, where I use this to prevent getting notified of NodeJS 18 (Current) where I want to stay on 16 (LTS).
Yet there might be a better solution to this I have usually opted to use this preset over that till now.

Anyway, I shall remove that preset from the list, as it is more a hindrance here.

.github/renovate.json
nscuro: Oh, I wasn't aware of the correlation with `engines`. In that case I think it's fine to keep it like it is. Thanks for the clarification!

.github/workflows/ci-release.yaml
nscuro: Ah alright. Maybe we should change the username and email to something that more directly indicates that a GH action did the commit: https://github.com/actions/checkout/issues/13#issuecomment-724415212

It looks like 
```suggestion
          git config user.name "github-actions[bot]"
          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```

does the trick.

k3rnelpan1c-dev: yeah, the values here where more a placeholder than anything real.

nscuro:
@k3rnelpan1c-dev @stevespringett I think we should decide on either Debian **or** Alpine. There's really no reason to have both. I'd personally vouch for Alpine - smaller attack surface and I've never seen issues with it. See also @k3rnelpan1c-dev's response here: https://github.com/DependencyTrack/frontend/issues/78#issuecomment-997400196

I'd rather not have the complexity of handling multiple flavors in CI if we can avoid it.

Thoughts?

stevespringett:
The API Server moved to Debian due to use of glibc and to avoid some issues that Alpine has with K8s.

The frontend we have a bit more flexibility over. I also prefer Alpine due to the smaller attack surface, but I need to find out what that K8s issue with Alpine was. But I would prefer to have a single base image rather than multiple.

.github/renovate.json
k3rnelpan1c-dev: Okay, will do.
However, on the note of keeping it consistent, the frontend repo seems to not have had Dependabot configured :?

k3rnelpan1c-dev:
I cannot imagine Alpine having any issue on K8s, after all the most commonly used external Ingress controller happens to be Nginx, which is using its Alpine based version IIRC.

But I am curious, if you happen to find the reference / remember what this issue might have been that would be great input. The only "issue" that I am aware with Alpine was that programs relying on glibc could experience very odd behaviour, however that was in every container environment not just K8s.

nscuro:
Known issues appear to evolve around DNS resolution. Small selection:
* https://github.com/kubernetes/kubernetes/issues/30215
* https://github.com/DependencyTrack/dependency-track/issues/1054
* https://github.com/kubernetes/kubernetes/issues/64924
* https://github.com/grafana/grafana/issues/20153

The K8s team themselves [claim](https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/):

> If you are using Alpine version 3.3 or earlier as your base image, DNS may not work properly due to a known issue with Alpine. Kubernetes [issue 30215](https://github.com/kubernetes/kubernetes/issues/30215) details more information on this.

 The frontend doesn't do any DNS resolution though. Even if the issue(s) above persist, it won't be a problem for this particular container.

k3rnelpan1c-dev:
took the liberty and removed the Debian flavour from the build and release process, as we all seem to be in agreement that we would prefer to only offer one flavour and that the attack surface of Alpine Linux is favourable.

nscuro:
@k3rnelpan1c-dev Apologies for my commits. Wanted to test in my fork but forgot to specify the correct remote when pushing my changes... May I ask you to hard-reset them, verify that everything's still as you intended and force push? 🥲

nscuro:
Anyway, I was now able to test this and it works like a charm!
* Build CI: https://github.com/nscuro/frontend/actions/runs/2316174514
* Build Release: https://github.com/nscuro/frontend/actions/runs/2316236491
* Release: https://github.com/nscuro/frontend/releases/tag/4.5.0
* Containers: https://github.com/nscuro/frontend/pkgs/container/frontend

Once we have my mess 🥲 cleaned up it's good to go from my side.

k3rnelpan1c-dev:
all done + added a tiny hotfix that we should consider adding to the backed CI eventually.
The login against `docker.io` should only really run when we are actually attempting to push the image to it (or any registry for that matter).

k3rnelpan1c-dev:
Thank you for testing this nscuro :) 
+ I could have locked the PR if I wanted to prevent upstream maintainers to change my changes, so no harm done in any way.

nscuro (approved):
All good from my side, just need @stevespringett to approve it before we can merge.

stevespringett (approved):

stevespringett:
Huge thanks @k3rnelpan1c-dev

nscuro:
It looks like, when releasing, `_meta-build.yml/build-node` does not "see" the version bump from `ci-release.yml/prepare-release`. `actions/checkout` per default checks out the Git commit that triggered the workflow run, which will always be the one *before* `npm version` was called in `ci-release.yml/prepare-release`.

For example, https://github.com/DependencyTrack/frontend/runs/6483307556?check_suite_focus=true checked out https://github.com/DependencyTrack/frontend/commit/01b253ce85ec3164714e6a2266b8c6024ad07733 instead of https://github.com/DependencyTrack/frontend/commit/3fe6bf7eefb34bbb8c635ca9a8fa2e09d50f49a6. 

This results in the `version` field of `package.json` not being correct, which also shows in the UI:
image

I'm wondering what would be the best strategy to resolve this. Some ideas:

* Decouple `prepare-release` from the rest of the workflow. Essentially, only run `npm version` (which creates a Git tag)
  * Introduce another workflow that's triggered by new tags. Run builds and `create-release` in that workflow
* Somehow make `_meta-build.yml` aware that the to-be-built commit changed. I feel like this may be wonky and error prone.

Would love to get your input @k3rnelpan1c-dev, your expertise is greatly appreciated.

k3rnelpan1c-dev:
Oha, well this is interesting and certainly unexpected behaviour :thinking: 
I will try and toy around with this the coming weekend and see whats up with this behaviour.

Anyway, for feedback on your workaround proposals: Both can work.
We could split off the build and publish part of the release CI and have it only create a GH Release that would trigger a build + publish CI or, give I can figure out why GHA behaves the way it does, we fix whatever causes it. (given we can, as it dose not seem to have that problem in the backend CI)

k3rnelpan1c-dev:
Okay I thought of this a bit more and it actually makes sense that by default every checkout step within the same workflow checks out the state that it got triggered with .... (and now I am annoyed that I didn't think of this earlier)

In other news I have a feeling that the idea of creating a pure "create release" workflow and letting a "publish" workflow trigger on "release creation" might be the right way to move forward. 

I will analyse more, but otherwise would commit to that idea and iterate over the CI that way.
Given you are on board with that idea, ofc.

k3rnelpan1c-dev:
Okay, tiny problem with the separated workflows: No matter how I change it using the default `GITHUB_TOKEN` cannot be used to realise it :frowning_face: 

> _Note on GitHub Actions:_ You can use the default token which is provided in  the secret _GITHUB_TOKEN_. However releases done with this token will NOT trigger release events to start other workflows.
> If you have actions that trigger on newly created releases, please use a generated token for that and store it in your repository's secrets (any other name than GITHUB_TOKEN is fine).

So to say that I am stuck with that idea is quite accurate, unless we opt to setup a "Dependency-Track-Bot" account that will serve as PAT donor to create the release and commits to the projects on release.

PR #144, opened 2022-04-29
Show Security Risk in policy violations overview (fixes #91)

### LINKED ISSUE #91: Policy Violations Breakdown Missing Security Policies
### Current Behavior:

DT v4.1 introduced support for vulnerabilities in policy violations.  In DT 4.3.1, the"Overview" tab for individual projects does not include Security Risk in the Policy Violations Breakdown chart:

image

This screenshot was taken from a project with 0 licence violations and 0 operational violations (ie, stats are correct) but which does have 2 security policy violations.

### Steps to Reproduce:
* Create a security policy such as `subject == SEVERITY && value IS MEDIUM` (or whatever can guarantee at least one violation) 
* For purpose of demonstration there is no need to restrict the policy to a particular project...  although this can be done in order to reduce "noise".
*  Navigate to a project that will be in violation of this policy
*  The Policy Violations tab on the project page will list violations of type "security".  Count them.
* The Overview tab on the project page displays a Policy Violations chart, broken down by Classification.  Observe that there is no info on security violations.

### Expected Behavior:
The Policy Violations by Classification Chart should include "security risk" and the number should match what you counted on the Policy Violations tab.

### Environment:
  
  - Dependency-Track Version: 4.3.1
  - Client Browser: Firefox 90.0
  - Client O/S: Windows 10.

### Additional Details:

I am guessing that the problem relates to the code in [ChartPolicyViolationBreakdown.vue](https://github.com/DependencyTrack/frontend/blob/master/src/views/dashboard/ChartPolicyViolationBreakdown.vue) that is commented out and marked TODO.

AbdelHajou:
@stevespringett Do you remember why this was commented out? The code seems to be working just fine:
Security Risk

Fixes #91

### DIFF
--- a/src/views/dashboard/ChartPolicyViolationBreakdown.vue
+++ b/src/views/dashboard/ChartPolicyViolationBreakdown.vue
@@ -1,6 +1,5 @@
 <template>
   <div>
-<!-- TODO
     <div class="progress-group">
       <div class="progress-group-header">
         <i class="fa fa-shield fa-fw progress-group-icon"></i>
@@ -11,7 +10,6 @@
         <b-progress height={} class="progress-xs" :value="securityPercent" variant="violation-type"></b-progress>
       </div>
     </div>
--->
     <div class="progress-group">
       <div class="progress-group-header">
         <i class="fa fa-balance-scale fa-fw progress-group-icon"></i>


stevespringett:
You know, I think it was likely that when the policy engine first came out, only license and operational risk were possible. The ticket was specific about license policy. Security policy was added in a follow-up release. I likely added all three and commented out the security portion because it wasn't available and forgot to uncomment it. That's likely what happened.

Thanks for identifying and fixing the issue.

PR #149, opened 2022-05-14
feat: support xhr cross-site cookie credentials

### LINKED ISSUE #156: Support xhr cross-site cookie credentials
Currently all requests to the API from the frontend are not sending cookie credentials when the api and frontend reside on different domains since the axios config option `withCredentials: true` for doing so is not enabled.

I have a use case currently where i'm deploying dependencytrack behind a SSO proxy that uses session cookies to manage authorization. Since axios and ajax are not configured to send credentials, all API requests fail with a 401 since the cookie is absent.

I'd like to introduce an environment variable that will grant users the ability to enable cross-site cookie credentials.

### Current Behavior:
- User navigates to dependencies.example.com
- SSO proxy is unable to authorize the user, so the user is redirect to their provider login
- Login is successful and a authorization cookie is created under example.com
- User tries to login using the frontend
- SSO proxy responds with a 401 since the authorization cookie is missing when a POST is sent to `/api/v1/user/login`

### Proposed Behavior:
- Add a new environment variable `API_WITH_CREDENTIALS` that allows toggling support for cross-site cookie credentials
- Cookies are sent in all requests when `API_WITH_CREDENTIALS` environment variable is set to true

### PR that introduces this feature
- https://github.com/DependencyTrack/frontend/pull/149

mohamed-cds:
Closes #156 

Currently all requests to the API from the frontend are not sending cookie credentials when the api and frontend reside on different domains since the axios config option `withCredentials: true` for doing so is not enabled.

I have a use case currently where i'm deploying dependencytrack behind a SSO proxy that uses session cookies to manage authorization. Since axios is not configured to send credentials, all API requests fail with a 401 since the cookie is absent.

I'd like to introduce an environment variable that will grant users the ability to enable cross-site cookie credentials.

#### Current
- User navigates to dependencies.example.com
- SSO proxy is unable to authorize the user, so the user is redirect to their provider login
- Login is successful and a authorization cookie is created under example.com
- User tries to login using the frontend
- SSO proxy responds with a 401 since the authorization cookie is missing when a POST is sent to `/api/v1/user/login`

#### Proposed change
- Cookies are sent in all requests when `API_WITH_CREDENTIALS` environment variable is set to true


nscuro:
Sorry for the long delay, but thanks for the PR @mohamed-cds!

nscuro (approved):

PR #150, opened 2022-05-16
Feature/jira gchat integration

elastic-pangolin:
additions as promised in Issue https://github.com/DependencyTrack/dependency-track/issues/1583


fnxpt:
Theres already a jira integration not sure if this makes sense

PR #164, opened 2022-06-05
Add support for private package repositories

AbdelHajou:
This front-end change adds username and password fields to the Repository creation form. This allows Dependency-Track to retrieve package versions from private package repositories. Back-end changes are in DependencyTrack/dependency-track#1687


nscuro (approved):
Thanks @AbdelHajou!

Issue #98, opened 2021-10-29
Bug : Cannot continue to tape the team's name if stopping for a few seconds


Whisper40:
When you stop tipping the team's name for a few seconds it saves the record, but you can't continue to tape it.
So you have to know the name of the team before tipping it.

The textfield should keep the input even after saving it
![Animation](https://user-images.githubusercontent.com/30440458/139394064-3edaf751-3b90-4cf8-aeab-a445810c8a09.gif)

Mvld3r:
This bug is also present in the following inputs:

1. OIDC group names (found in `Administration > Access Management > OpenID Connect Groups > Select a existing Group`)
2. Repository URLs (found in `Administration > Repositories > Select a repository group > Select a existing repository`)
3. Policy names (found in `Policy Management > Select a existing policy`)
4. License group names (found in `Policy Management > License Groups tab > Select a existing policy group`)

PR #173, opened 2022-06-16
Resume focus after saving

Mvld3r:
Addresses https://github.com/DependencyTrack/frontend/issues/98

Signed-off-by: ege-dw <enora.germond@deveryware.com>


nscuro (approved):
Thanks for the PR @Mvld3r!

PR #249, opened 2022-09-18
Remove unnecessary hz-scrollbar on expanded rows

### LINKED ISSUE #248: Unnecessary horizontal scrollbar shown in expanded rows
### Current Behavior:

The frontend is showing a unnecessary horizontal scrollbar when a section inside a table is expanded. The scrollbar scrolls only a few pixels, which is unnecessary because all of the content fits on the screen.

grafik

### Steps to Reproduce:

Go to some place where a section in a table is expandable, which is a common pattern on many views of the frontend. Expand it, scroll down and search the horizontal scrollbar. Places to go:

- Administration, Notifications, Alerts. Then expand an alert in the list.
- Project, Audit Vurnerabilities. Then expand a vurnerability in the list, the scrollbar will show up on the bottom of the screen.
- And other places with same ui pattern.

### Expected Behavior:

The frontend should not show a horizontal scrollbar until it's necessary.

### Environment:
  
  - Dependency-Track Version: v4.6.0-SNAPSHOT (frontend master)
  - Client Browser: Chrome 105, Firefox 104
  - Client O/S: Windows 11

### Additional Details:

Because of the styling this defect can be seen most clearly on Chrome, but happens on Firefox as well. I haven't test it on Safari.

tmehnert:
This effectively reduce the width of an `expanded-row` by 2 pixels to remove the scrollbar, when it is not necessary. It turns out that `border-collapse: collapse;` on the table causing the scrollbars to be shown. Because of setting this attribute to `seperate` will change the appearance, i adjusted the margin of the expanded-rows instead. The commit will overwrite the `margin-right: -15px`, which is set by the row class.

Tested on Chrome and Firefox. Please someone test it on Safari.

Fixes #248

### DIFF
--- a/src/assets/scss/_custom.scss
+++ b/src/assets/scss/_custom.scss
@@ -227,6 +227,7 @@ input.required, select.required, textarea.required {
 .expanded-row {
   margin-top:-12px;
   margin-bottom:-12px;
+  margin-right: -13px;
   padding-top:20px;
   padding-bottom: 20px;
   background-color: $grey-800;


nscuro (approved):
Thanks for that @tmehnert, attention to detail truly matters!

Tested on Safari BTW, it works there as well.

PR #251, opened 2022-09-19
Fix Project View Details display wrong tags

### LINKED ISSUE #188: Project View Details Can Display Wrong Tags
### Current Behavior:
After adding or removing tags from version X of project A, and then switching via dropdown to version Y of project A,  the tags for Y look fine...  but clicking "project details" lists the tags for version X.

### Steps to Reproduce:

- Navigate to any project that has 2 or more versions.  ie,version X of project A
- Click on "view details" for this version X and add and remove some tags (make a note of what the tags were, so you can restore them later!)
- Click Updateto save your tag changes
- Use dropdown to switch to version Y of project A
- Click on "view details" for this version Y and note that the tags match what you just configured for version X.  They are not the correct tags for Y.

### Expected Behavior:
Tags in "view details" match the actual tags on the project.

### Environment:
  
  - Dependency-Track Version: 4.5.0 (and also in 4.6.0 SNAPSHOT)
  - Client Browser:  Firefox 102
  - Client O/S: Windows 10

### Additional Details:
I tried to see if the same behaviour occurred for description and type.  It did  not.  All was OK with those.

tmehnert:
This will fix a bug that cause the wrong tags to be displayed in the details dialog when switching between projects versions. For more details about the impact, see linked issue. This caused by a conversation routine between DTO and Domain-Model, which was skipped when the DTO was allready initialized. In this case the previously displayed tags were displayed, even if they where not defined in the version of the project.

Fixes #188

### DIFF
--- a/src/views/portfolio/projects/ProjectDetailsModal.vue
+++ b/src/views/portfolio/projects/ProjectDetailsModal.vue
@@ -1,5 +1,5 @@
 <template>
-  <b-modal id="projectDetailsModal" size="md" hide-header-close no-stacking :title="$t('message.project_details')">
+  <b-modal id="projectDetailsModal" size="md" hide-header-close no-stacking :title="$t('message.project_details')" @show="initializeTags">
     <b-tabs class="body-bg-color" style="border:0;padding:0">
       <b-tab class="body-bg-color" style="border:0;padding:0" active>
         <template v-slot:title><i class="fa fa-edit"></i> {{ $t('message.general') }}</template>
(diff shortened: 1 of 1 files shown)


tmehnert:
Codacy found an issue in my code, i will try to understand it.

tmehnert:
Okay changed my approach, now the tags bound to the input field are initialized when the modal is shown.

nscuro (approved):

nscuro:
Thanks for the PR @tmehnert! 🚀

PR #254, opened 2022-09-20
Add missing translation calls

tmehnert:
This is a pure i18n hardcode-cleanup, because of none of these affects the user experience, i didn't create an issue. I replaced some hardcoded texts with proper calls to i18n. Also set texts for b-file-form because this ui-component seems to not have an own translation in styled mode. Some missing translations, which i found, i have left as is.

**Tooltip in EPSS Chart**
I didn't know how to access i18n in the [label callback of ChartEpssVsCvss](https://github.com/DependencyTrack/frontend/blob/1c2484271996b62104a02c3022f2fffef654b68b/src/views/dashboard/ChartEpssVsCvss.vue#L92). If someone can give me a hint, i can add it.

**Texts in controls of the table (bootstrap-table)**
The bootstrap-table ui-component has it's own maintained translation for tooltips or the pagination state for example. These are stored as Javascript files which need to be included. But not for english, so i skipped this.


nscuro (approved):
Thanks @tmehnert!

> I didn't know how to access i18n in the [label callback of ChartEpssVsCvss](https://github.com/DependencyTrack/frontend/blob/1c2484271996b62104a02c3022f2fffef654b68b/src/views/dashboard/ChartEpssVsCvss.vue#L92). If someone can give me a hint, i can add it.

I just tested it, and it looks like you can just add this to the top of `ChartEpssVsCvss.vue`:

```javascript
import i18n from "../../i18n";
```

and then whenever you need to use it, just call 

```javascript
i18n.t('message.<IDENTIFIER>')`
```

tmehnert:
@nscuro translation in EPSS Chart added.

PR #271, opened 2022-10-04
Allow selection of teams as recipients for alert rules

rbt-mm:
Improvement as suggested in #1608.
Also partly adresses #1601

Managing the destination of a notification rule takes a lot of time for a large group of recipients because it needs to be done manually by a user for every single recipient.

Allowing teams to be subscribed to an email alert automates this process, as every user of a subscribed team will automatically get an email without the need of inserting their email address as a destination.

Not having to manually insert every single email address into the notification rule destination simplifies the management of notification recipients and allows dynamic email destinations by subscribing and reorganizing teams.
Individual persons can still be added as recipients if they are manually put as a destination.

[Backend PR](https://github.com/DependencyTrack/dependency-track/pull/2008)

Signed-off-by: RBickert [rbt@mm-software.com](mailto:rbt@mm-software.com)


src/views/administration/notifications/Alerts.vue
```diff
@@ -134,8 +135,18 @@
                       <b-form-input id="input-3" v-model="notificationLevel" disabled class="form-control disabled" readonly trim />
                     </b-form-group>
                     <b-input-group-form-input id="input-destination" :label="$t('admin.destination')" input-group-size="mb-3"
-                                              required="true" type="text" v-model="destination" lazy="true"
+                                              :required="(!(this.alert.hasOwnProperty('teams') && this.alert.teams != null && this.alert.teams.length > 0)).toString()"
+                                              type="text" v-model="destination" lazy="true"
                                               v-debounce:750ms="updateNotificationRule" :debounce-events="'keyup'" />
+                     <b-form-group v-if="this.publisherClass === 'org.dependencytrack.notification.publisher.SendMailPublisher'"
```
nscuro, line 141: This block is currently displayed for scope `SYSTEM` too, despite addition of teams to `SYSTEM` alerts not being possible according to the API server code. 

I had a related question in the API server PR, probably makes sense to clarify that first.

nscuro (commented):
Looks good, too! Only handling of scope needs clarification.

src/views/administration/notifications/Alerts.vue
nscuro: Considering this as solved, as the scope limitation has been removed on the API server side.

nscuro (approved):

Issue #84, opened 2021-08-01
Vulnerabilities: Filter out where "Projects = 0"


msymons:
### Current Behavior:
With a totally fresh install of Dependency-Track 4.2.2, the Vulnerability page very quickly "fleshes out" to display more than 125k rows.  This provides a usability issue with getting an overall view of "what is affecting me".

### Proposed Behavior:
Provide a checkbox similar to the " Show inactive projects" displayed on the Projects screen.  The checkbox would act to include/exclude all vulnerabilities that have 0 affected projects, with the suggested default being to exclude.  This would reduce the displayed listing from 125k to a (hopefully) a nice low number.  Even if there 1000 vulnerabilities it would still only take 10 clicks to navigate from beginning to end with display set to 100 per page.

One use case for this suggested functionality is that it would make it possible to sort vulnerabilities by "Published" and then simply scroll down to see the most recent vulnerabilities to appear in the portfolio.  Useful should notifications be unconfigured (or directed to the wrong people, etc).

sekwah41:
If bootstrap table is updated to 1.19.1 custom buttons are supported on the tables.

Though this may require a backend change to support the query. If this is possible to filter would someone be able to supply the required parameter for the api/v1/vulnerability endpoint and I can have a go at making some proposed UI for applying additional features with some quick buttons for parameters like this.

stevespringett:
Changes to the backend would need to be made. Specifically, a check to see if the vulnerability contains components or services. If either one are true, then the vulnerability would be part of the DB response.

terbolous:
To chime in, it would also be nice to be able to sort on the number of projects affected

sekwah41:
I've had a pr merged now which would make the UI changes possible. The old version didn't support custom buttons and I spent a good hour trying to debug why they wouldn't even begin to show before seeing the documentation was outdated :sweat_smile:

PR #272, opened 2022-10-04
Show hierarchical view in project list

rbt-mm:
Backend issue [#84](https://github.com/DependencyTrack/dependency-track/issues/84)

Added a switch in the project list to either display a hierarchical view, which shows the (already existing) multi-level parent-child-relationships between projects, or to display the currently existing flat view of projects.

Added new select inputs in the dialogs for creating a project or showing its details, which include every possible parent for the current or new project and create a persistent relationship between them on selection.

Introducing new parent-child relationship logic:

- Prevents that inactive projects can be selected as parents
- A project cannot be set to inactive if it has an active child
- A parent project cannot select one of his children as a parent

Create Project
Project Details
Project List Flat View
Project List Hierarchical View Collapsed
Project List Hierarchical View Expanded

[Backend PR](https://github.com/DependencyTrack/dependency-track/pull/2009)

Signed-off-by: RBickert [rbt@mm-software.com](mailto:rbt@mm-software.com)


rkg-mm:
@nscuro I'm a bit clueless what the issue raised by static code analysis wants to tell us or how to improve it, but we never worked with Vue before. Any idea how to fix it?

src/views/portfolio/projects/ProjectList.vue
```diff
@@ -27,17 +28,81 @@
...
+            onExpandRow: this.vueFormatterInit
+          }
+        }
+      }
+
+    }
+  }
+
+  function detailFilter(index, row){
+    return (Object.prototype.hasOwnProperty.call(row, 'children') && row.children)
```
nscuro, line 85: This is currently causing the expand button and an empty table to be displayed, in cases where no child is active:

Screenshot from 2022-10-18 14-06-53


This appears to resolve it:

```suggestion
    return (Object.prototype.hasOwnProperty.call(row, 'children') && row.children && row.children.some(child => child.active));
```

nscuro (approved):

stevespringett:
@rbt-mm Thanks for the PR. 

@nscuro I really do not like how this was implemented. The whole table within table presentation is hard to look at. I would have expected the treegrid plugin for bootstrap table to be used, but was not. Early prototypes were available that did this and by doing so, you'll avoid having to have tables within tables.

Let's rethink this. I'd rather back out this merge and use treegrid for all parent/child relationships in a table structure.

https://bootstrap-table.com/docs/extensions/treegrid/

stevespringett:
Diff attached
[parent-child_projects.patch.zip](https://github.com/DependencyTrack/frontend/files/9812412/parent-child_projects.patch.zip)

nscuro:
@stevespringett Oh wow that indeed does look a lot better. I will revert the merge.

stevespringett:
@rkg-mm It was a prototype, so barely working, but it did work.

Pagination? Yes and no. Pagination works on the table as usual, but I don't want to support paginated tables within paginated tables. Thats confusing.

lazy-loading sub-levels? No, not supported, but theoretically it could be, and likely should be.

rkg-mm:
* Pagination I didn't mean within sub-levels, just the root level (also needed for people that do not use the hierarchy I think).
* I also don't find a reference for lazy loading. Would need to find a way to make it work. Possibly using the "expanded"-event and some tricks to get a loading indicator into the sublevel while its loading...

rkg-mm:
@stevespringett mind sharing your prototype for a quick start?

rkg-mm:
also, would you still want to keep both options (flat + hierarchy)? or only the tree-view then?

stevespringett:
@rkg-mm The prototype was posted in https://github.com/DependencyTrack/frontend/pull/272#issuecomment-1282631260

rkg-mm:
@stevespringett  yes thanks, had overseen it but my colleague told me later :)

PR #286, opened 2022-10-17
Issue-365 : Snyk integration frontend

sahibamittal:
**Issue:** https://github.com/DependencyTrack/dependency-track/issues/365

Screenshot 2022-10-17 at 5 20 17 p m


src/shared/common.js
```diff
@@ -118,6 +118,10 @@ $common.formatAnalyzerLabel = function formatAnalyzerLabel(analyzer, vulnSource,
       analyzerLabel = "VulnDB";
       analyzerUrl = "https://vulndb.cyberriskanalytics.com/vulnerabilities/" + vulnId;
       break;
+    case 'SNYK_ANALYZER':
+      analyzerLabel = "Snyk";
+      analyzerUrl = "https://security.snyk.io/vuln/" + vulnId;
```
nscuro, line 123: This needs to be added in the `case 'INTERNAL_ANALYZER'` branch a few lines above as well, when `vulnSource == 'SNYK'`. 

Because the API server is mirroring vulnerable versions, snyk vulns may be identified by the internal analyzer, too. In that case, there's no reference URL to Snyk rendered in the UI right now:
image

Versus when the vulnerabilities are first identified by the Snyk analyzer:
image

src/views/portfolio/vulnerabilities/Vulnerability.vue
```diff
@@ -174,6 +174,8 @@
             return "Sonatype OSS Index";
           case 'VULNDB':
             return "VulnDB (Risk Based Security)";
+          case 'SNYK':
+            return "Snyk Vulnerability";
```
nscuro, line 178: The "Vulnerability" here seems redundant.

nscuro (changes requested):

nscuro (approved):

PR #315, opened 2022-11-04
aliases: improve alias column, show aliases optionally in all vulnerability lists 

valentijnscholten:
### Description

2023-02-15 Attempt 2 at this PR: 

The Audit Vulnerabilities tab for a project has an extra column that can be shown to list the aliases of a vulnerability. This PR improves the alignment of these aliases to make it look a little bit better.

Old:

image

New:

image

I have also added this optional column to the global Vulnerabilities and Components lists

### Addressed Issue

- Vulnerabilities with multiple aliases look ugly in Audit Vulnerabilities tab
- Alias column can be useful also in Components list and Vulnerabilities list

### Additional Details

Not everybody might feel this is the "perfect" solution, but let's not [make "perfect" the enemy of "good"](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good) :-)

Not sure if there's a practical way to wrap around each alias without using <br/>

### Checklist

Signed-off-by: Valentijn Scholten <valentijnscholten@gmail.com>

### DIFF
--- a/src/views/portfolio/projects/ComponentVulnerabilities.vue
+++ b/src/views/portfolio/projects/ComponentVulnerabilities.vue
@@ -39,6 +39,24 @@
             }
           },
           {
+            title: this.$t('message.aliases'),
+            field: "aliases",
+            sortable: true,
+            visible: false,
+            formatter(value, row, index) {
+              if (typeof value !== 'undefined') {
+                let label = "";
+                for (let i=0; i<value.length; i++) {
+                  let alias = common.resolveVulnAliasInfo(row.source, value[i]);
+                  let url = xssFilters.uriInUnQuotedAttr("../vulnerabilities/" + alias.source + "/" + alias.vulnId);
+                  label += common.formatSourceLabel(alias.source) + ` <a href="${url}">${xssFilters.inHTMLData(alias.vulnId)}</a>`
+                  if (i < value.length-1) label += "<br/><br/>"
+                }
+                return label;
+              }
+            }
+          },
+          {
             title: this.$t('message.published'),
             field: "published",
             sortable: true,
--- a/src/views/portfolio/projects/ComponentVulnerabilities.vue
+++ b/src/views/portfolio/projects/ComponentVulnerabilities.vue
@@ -41,7 +41,6 @@
           {
             title: this.$t('message.aliases'),
             field: "aliases",
-            sortable: true,
             visible: false,
             formatter(value, row, index) {
               if (typeof value !== 'undefined') {
--- a/src/views/portfolio/vulnerabilities/VulnerabilityList.vue
+++ b/src/views/portfolio/vulnerabilities/VulnerabilityList.vue
@@ -64,7 +64,6 @@
           {
             title: this.$t('message.aliases'),
             field: "aliases",
-            sortable: true,
             visible: false,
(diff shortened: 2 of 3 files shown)


nscuro:
Thanks for the PR @valentijnscholten! 

IIRC this was deemed to be a non-starter in the [related Slack discussion](https://owasp.slack.com/archives/C6R3R32H4/p1667585442465429). So I think this can be closed until we come up with a better way to visualize this data.

valentijnscholten:
It's already used in the Project Findings list, so why not just copy it into the other places for now so we gain some usability around aliases?

valentijnscholten:
See: https://owasp.slack.com/archives/C6R3R32H4/p1667585442465429

valentijnscholten:
Made some small changes to improve the layout to the existing column already available since 4.6.0 and updated the description to hopefully convince this would be a good PR to merge.

src/views/portfolio/vulnerabilities/VulnerabilityList.vue
```diff
@@ -61,6 +61,24 @@
               return common.formatSourceLabel(row.source) + ` <a href="${url}">${xssFilters.inHTMLData(value)}</a>`;
             }
           },
+          {
+            title: this.$t('message.aliases'),
+            field: "aliases",
+            sortable: true,
```
stevespringett, line 67: I don't think aliases are sortable, are they?

src/views/portfolio/projects/ComponentVulnerabilities.vue
```diff
@@ -38,6 +38,24 @@
               return common.formatSourceLabel(row.source) + ` <a href="${url}">${xssFilters.inHTMLData(value)}</a>`;
             }
           },
+          {
+            title: this.$t('message.aliases'),
+            field: "aliases",
+            sortable: true,
```
stevespringett, line 44: I don't think aliases are sortable, are they?

src/views/portfolio/vulnerabilities/VulnerabilityList.vue
valentijnscholten: Good point. They were marked as sortable in the existing alias column in ProjectFindings.vue which was copied over. But it doesn't work correctly, so i've disabled sorting in all 3 places now.

src/views/portfolio/projects/ComponentVulnerabilities.vue
valentijnscholten: Good point. They were marked as sortable in the existing alias column in ProjectFindings.vue which was copied over. But it doesn't work correctly, so i've disabled sorting in all 3 places now.

valentijnscholten:
I also looked at a "new line" solution without using <br/>. I see a lot of people using `<pre>` (or similar) requiring `\n` in our html to achieve the same. Some others use <p>. I'm oldschool, so I left the <br/> in place.

valentijnscholten:
Haha the BR tag is rendered as html by GitHub

valentijnscholten:
@stevespringett Do you want me to set the CWE column to not sortable as well? It behaved erratically, even if all rows have only 1 CWE.

image

BTW In other places the CWE column is rendered differently (and not sortable):

image

nscuro (approved):

Issue #168, opened 2022-06-15
Display added/modified project tags in lowercase after validation


Mvld3r:
This ER relates to https://github.com/DependencyTrack/dependency-track/issues/1717, which requests to make search projects by tags case-insensitive, and is a follow-up in the frontend should https://github.com/DependencyTrack/dependency-track/issues/1717 be accepted.

### Current Behavior:
When a tag is added to (or modified in) a project (for example, Projects > create Project), after validating the tag (by pressing 'enter' or 'space'), the vuejs tag preserves the case-sensitivity that the project tag was input in.

### Proposed Behavior:
Should the project search by tag become case-insensitive (https://github.com/DependencyTrack/dependency-track/issues/1717), a validated tag should be displayed in lowercase to reflect how it will be saved in the backend.

Resulting behaviour:
![lowercase_tag](https://user-images.githubusercontent.com/26394852/173865090-d076b0b9-4091-4cb2-bef6-4ba122d606d3.gif)

PR #319, opened 2022-11-05
project list: show tags

valentijnscholten:
Currently tags are not shown in the project list. This PR adds them in a basic way in an optional column:

image

Signed-off-by: Valentijn Scholten <valentijnscholten@gmail.com>

### DIFF
--- a/src/views/portfolio/projects/ProjectList.vue
+++ b/src/views/portfolio/projects/ProjectList.vue
@@ -185,7 +185,14 @@
             sortable: true,
             visible: false,
             formatter(value, row, index) {
-              return row.tags?.map(tag => `<a href="../projects/?tag=${xssFilters.uriComponentInUnQuotedAttr(tag)}">${xssFilters.inHTMLData(tag)}</a>`).join(', ') || '';
+              let tag_string = row.tags.slice(0, 2)?.map(tag => common.formatProjectTagLabel(tag)).join(', ') || '';
+              if (row.tags.length > 2) {
+                tag_string += `<span class="d-none">`
+                  tag_string += ', ' + row.tags.slice(1)?.map(tag => common.formatProjectTagLabel(tag)).join(', ');
+                  tag_string += `</span>`
+                  tag_string += `<a href="#" title="show all tags" class="badge badge-tag" onclick="this.previousElementSibling.classList.toggle('d-none')">…</a>`
+              }
+              return tag_string;
             }
           },
           {
--- a/src/views/portfolio/projects/ProjectList.vue
+++ b/src/views/portfolio/projects/ProjectList.vue
@@ -175,9 +175,8 @@
             field: "name",
             sortable: true,
             formatter(value, row, index) {
-              let name_url = xssFilters.uriInUnQuotedAttr("../projects/" + row.uuid);
-              let projectName = `<a href="${name_url}">${xssFilters.inHTMLData(value)}</a>`;
-              return projectName
+              let url = xssFilters.uriInUnQuotedAttr("../projects/" + row.uuid);
+              return `<a href="${url}">${xssFilters.inHTMLData(value)}</a>`;
             }
           },
           {
@@ -186,19 +185,7 @@
             sortable: true,
             visible: false,
             formatter(value, row, index) {
-              if (!row.tags) {
-                return "";
-              } else {
-                let tagNames = ""
-                  for (let i=0; i<row.tags.length; i++) {
-                  let tag = row.tags[i].name
-                  let tag_url = xssFilters.uriInUnQuotedAttr("../projects/?tag=" + tag);
-                  let tagName = `<a href="${tag_url}" target="_self" class="">${tag}</a>`
-                  tagNames += tagName
-                  if (i < row.tags.length-1) tagNames += ", "
-                }
-                return tagNames
-              }
+              return row.tags?.map(tag => `<a href="../projects/?tag=${xssFilters.uriComponentInUnQuotedAttr(tag)}">${xssFilters.inHTMLData(tag)}</a>`).join(', ') || '';
             }
           },
           {
--- a/src/views/portfolio/projects/ProjectList.vue
+++ b/src/views/portfolio/projects/ProjectList.vue
@@ -182,7 +182,7 @@
           {
             title: this.$t('message.tags'),
             field: "tags",
-            sortable: true,
+            sortable: false,
             visible: false,
             formatter(value, row, index) {
               let tag_string = row.tags.slice(0, 2)?.map(tag => common.formatProjectTagLabel(tag)).join(' ') || '';
--- a/src/views/portfolio/projects/ProjectList.vue
+++ b/src/views/portfolio/projects/ProjectList.vue
@@ -170,7 +170,7 @@
           dataOff: '\u2715'
         },
         columns: [
-        {
+          {
             title: this.$t('message.project_name'),
             field: "name",
             sortable: true,
--- a/src/views/portfolio/projects/ProjectList.vue
+++ b/src/views/portfolio/projects/ProjectList.vue
@@ -187,8 +187,8 @@
             formatter(value, row, index) {
               let tag_string = row.tags.slice(0, 2)?.map(tag => common.formatProjectTagLabel(tag)).join(', ') || '';
               if (row.tags.length > 2) {
-                tag_string += `<span class="d-none">`
-                  tag_string += ', ' + row.tags.slice(1)?.map(tag => common.formatProjectTagLabel(tag)).join(', ');
+                tag_string += `, <span class="d-none">`
+                  tag_string += row.tags.slice(1)?.map(tag => common.formatProjectTagLabel(tag)).join(', ');
                   tag_string += `</span>`
                   tag_string += `<a href="#" title="show all tags" class="badge badge-tag" onclick="this.previousElementSibling.classList.toggle('d-none')">…</a>`
               }
--- a/src/views/portfolio/projects/ProjectList.vue
+++ b/src/views/portfolio/projects/ProjectList.vue
@@ -185,10 +185,10 @@
             sortable: true,
             visible: false,
             formatter(value, row, index) {
-              let tag_string = row.tags.slice(0, 2)?.map(tag => common.formatProjectTagLabel(tag)).join(', ') || '';
+              let tag_string = row.tags.slice(0, 2)?.map(tag => common.formatProjectTagLabel(tag)).join(' ') || '';
               if (row.tags.length > 2) {
-                tag_string += `, <span class="d-none">`
-                  tag_string += row.tags.slice(2)?.map(tag => common.formatProjectTagLabel(tag)).join(', ');
+                tag_string += ` <span class="d-none">`
+                  tag_string += row.tags.slice(2)?.map(tag => common.formatProjectTagLabel(tag)).join(' ');
                   tag_string += `</span>`
                   tag_string += `<a href="#" title="show all tags" class="badge badge-tag" onclick="this.previousElementSibling.classList.toggle('d-none')">…</a>`
               }
--- a/src/views/portfolio/projects/ProjectList.vue
+++ b/src/views/portfolio/projects/ProjectList.vue
@@ -175,8 +175,23 @@
             field: "name",
             sortable: true,
             formatter(value, row, index) {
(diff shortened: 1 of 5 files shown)


nscuro:
cc @msymons, because I know he's been wanting more features around tag management and visibility. Mark, any suggestions how you'd like tags to be displayed here?

msymons:
Should they perhaps be displayed in a separate (optional) column?  That way, the user has the choice about whether or not to display them.  It might also aleviate concerns about tags consuming too much screen real estate  (some of my projects have a lot more than the 3 tags used in the screenshot above).

Apart from that,  the concept of the PR is useful.  If the projects in the screenshot were in my DT, I would immediately be wanting to work out why some projects have no tag at all (and then fix them).

rkg-mm:
I agree, I would put them in a separate column probably, too

valentijnscholten:
Updated the PR to use an extra optional column.

src/views/portfolio/projects/ProjectList.vue
```diff
@@ -168,13 +168,35 @@
...
+        {
             title: this.$t('message.project_name'),
             field: "name",
             sortable: true,
             formatter(value, row, index) {
-              let url = xssFilters.uriInUnQuotedAttr("../projects/" + row.uuid);
-              return `<a href="${url}">${xssFilters.inHTMLData(value)}</a>`;
+              let name_url = xssFilters.uriInUnQuotedAttr("../projects/" + row.uuid);
+              let projectName = `<a href="${name_url}">${xssFilters.inHTMLData(value)}</a>`;
+              return projectName
```
sephiroth-j, line 178: I'll take the liberty of making a few comments on this, even though I'm only a guest here.

The change is completely unnecessary. Also, the extra variable assignment before the return is unnecessary.

src/views/portfolio/projects/ProjectList.vue
```diff
@@ -168,13 +168,35 @@
...
+              } else {
+                let tagNames = ""
+                  for (let i=0; i<row.tags.length; i++) {
+                  let tag = row.tags[i].name
+                  let tag_url = xssFilters.uriInUnQuotedAttr("../projects/?tag=" + tag);
+                  let tagName = `<a href="${tag_url}" target="_self" class="">${tag}</a>`
+                  tagNames += tagName
+                  if (i < row.tags.length-1) tagNames += ", "
+                }
+                return tagNames
```
sephiroth-j, line 198: There are several comments here.


1. the indentations are different or not correct (line 191).
2. missing semicolons
3. if statement without curly braces
4. used `let` where `const` would be possible
5. **a possible persistent XSS vulnerability in line 194 because `tag` is not escaped there!**
6. the tag-value in the url is not url-encoded. For example the tag "=foo" leads to an invalid URL.

I would recommend this.
```js
return row.tags?.map(tag => `<a href="../projects/?tag=${xssFilters.uriComponentInUnQuotedAttr(tag)}">${xssFilters.inHTMLData(tag)}</a>`).join(', ') || '';
```

src/views/portfolio/projects/ProjectList.vue
valentijnscholten: agreed (this was just a leftover from a previous iteration where the tags where included in the project name column)

src/views/portfolio/projects/ProjectList.vue
valentijnscholten: I think in a converstaion somewhere we concluded escaping should be done automatically here by bootstrap, but you're right, we need to do it ourselves. For the rest I am sometimes tempted to just copy and paste some code from other places and leave it in "imperative" style as it's easier to read.

but I've applied your oneline now.

valentijnscholten:
@sephiroth-j any suggestions to make the `tags` column more visually appealing?

sephiroth-j:
The following css classes on the `a`-tag would be good: "badge badge-tag text-uppercase mr-2". This would give them the same look as on the project page.

I would also limit the number of visible tags to a configurable number. If there are more tags, then show a "Show more" button (with "…" as text so it uses less space). Clicking the button will a) show the remaining tags and b) hide the button. The remaining tags can be wrapped in a `span`-tag which is hidden by default (css class "d-none"). The click on "show more" just toggles its visibility by removing the css class.

**simplified demo**
```html
<a class="badge badge-tag text-uppercase mr-2" href="#">tag1</a>
<a class="badge badge-tag text-uppercase mr-2" href="#">tag2</a>
<a href="#" title="show all tags" class="badge badge-tag" onclick="this.classList.toggle('d-none');this.nextElementSibling.classList.toggle('d-none')">…</a>
<span class="d-none">
	<a class="badge badge-tag text-uppercase mr-2" href="#">tag3</a>
	<a class="badge badge-tag text-uppercase mr-2" href="#">tag4</a>
</span>
```

valentijnscholten:
I can try that again, but in a previous iteration I had those classes and it didn't look to good.

I can also try the suggestion to limit the number of tags show and add a show all toggle. But will check how that affects search / ctrl-f behaviour.

valentijnscholten:
@sephiroth-j Your suggestion compiles into something that doesn't compile :-)

```
Module parse failed: Unexpected token (232:26)
File was processed with these loaders:
 * ./node_modules/cache-loader/dist/cjs.js
 * ./node_modules/babel-loader/lib/index.js
 * ./node_modules/cache-loader/dist/cjs.js
 * ./node_modules/vue-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|         visible: false,
|         formatter: function formatter(value, row, index) {
>           return row.tags?.map(function (tag) {
|             return "<a href=\"../projects/?tag=".concat(xssFilters.uriComponentInUnQuotedAttr(tag), "\">").concat(xssFilters.inHTMLData(tag), "</a>");
|           }).join(', ') || '';
```

I am no Vue/FrontEnd person and sometimes prefer an imperatie programming style which makes it easier to understand what's going on.

sephiroth-j:
This is just standard JavaScript, supported by every current browser. The build setup with Babel is quite outdated and hasn't been updated in years. Node.js 14 will be EOL at the end of April. Node.js 16, the old LTS, will be EOL this September. With Node.js 18, the build fails completely.

Anyway, either you rewrite it as legacy code with a (short) if-statement or you add [the optional chaining plugin](https://babel.dev/docs/babel-plugin-proposal-optional-chaining) to Babel.

```diff
diff --git a/babel.config.js b/babel.config.js
index b074cad..00abcf8 100644
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,4 +1,7 @@
 module.exports = {
+  plugins: [
+    "@babel/plugin-proposal-optional-chaining"
+  ],
   presets: [
     ['@vue/app', {
       polyfills: [
diff --git a/package.json b/package.json
index dabc722..70bb1f0 100644
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "xss-filters": "1.2.7"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-optional-chaining": "^7.21.0",
     "@cyclonedx/webpack-plugin": "2.0.0",
     "@vue/cli-plugin-babel": "3.12.1",
     "@vue/cli-plugin-eslint": "3.12.1",
```

valentijnscholten:
Styling options:

No styling:
image

Lowercase badges:
image


Uppercase badges:
image


@nscuro What do you think? My preference is no styling as the badges look to clumsy.

sephiroth-j:
> @nscuro Any idea why the dependency review check is failing? I just ran `npm install @babel/plugin-proposal-optional-chaining` and committed the resulting `package(-lock).json`.

I guess that's why.

Vulnerabilities
  package-lock.json » json5@2.2.1 – Prototype Pollution in JSON5 via Parse Method (high severity)
    ↪ https://github.com/advisories/GHSA-9c47-m6qq-7p4h
  Error: Dependency review detected vulnerable packages.

valentijnscholten:
> > @nscuro Any idea why the dependency review check is failing? I just ran `npm install @babel/plugin-proposal-optional-chaining` and committed the resulting `package(-lock).json`.
> 
> I guess that's why.
> 
> Vulnerabilities package-lock.json » json5@2.2.1 – Prototype Pollution in JSON5 via Parse Method (high severity) ↪ [GHSA-9c47-m6qq-7p4h](https://github.com/advisories/GHSA-9c47-m6qq-7p4h) Error: Dependency review detected vulnerable packages.

Ah, was blinded by the large output of the dependency review step. I updated `json5` just now.

nscuro:
@valentijnscholten I really like the expansion functionality (thx @sephiroth-j for the suggestion!). 

Of the visual styles you suggested I like option 2 the most (lowercase badges), but I'd remove the commas so it is uniform with how tags are displayed on the project page:

image
 
The tags column should also be marked as non-sortable.

valentijnscholten:
Should we go for lowercase badges in other places such as the Project view as well?

sephiroth-j:
> @sephiroth-j Thanks for the suggestions, I have implemented them. I chose to leave the '...' in place so people can chose to "show less" so to speak. Can you help me with 1 problem? I extract the formatting of the `a` for the tag to the `common` module. But as soon as I use `xssFilters` in there, it breaks everything. According to vscode it's OK and npm happily serves it, but with some warnings. In the browser I can see an error where the common module is trying to export its functions.

The use of `require` in `common.js` can not be used together with `import` in `ProjectList.vue`. To make it work, use `import` and `export default` in `common.js` as shown below.

```diff
diff --git a/src/shared/common.js b/src/shared/common.js
index abda651..5847344 100644
--- a/src/shared/common.js
+++ b/src/shared/common.js
@@ -6,7 +6,7 @@
 
 "use strict";
 
-const xssFilters = require("xss-filters");
+import xssFilters from "xss-filters";
 const $common = function() {
 };
 
@@ -463,7 +463,7 @@ $common.OWASP_RR_LIKELIHOOD_TO_IMPACT_SEVERITY_MATRIX = {
   }
 }
 
-module.exports = {
+export default {
   formatSourceLabel: $common.formatSourceLabel,
   formatNotificationLabel: $common.formatNotificationLabel,
   formatProjectTagLabel: $common.formatProjectTagLabel,
```

valentijnscholten:
> > @sephiroth-j Thanks for the suggestions, I have implemented them. I chose to leave the '...' in place so people can chose to "show less" so to speak. Can you help me with 1 problem? I extract the formatting of the `a` for the tag to the `common` module. But as soon as I use `xssFilters` in there, it breaks everything. According to vscode it's OK and npm happily serves it, but with some warnings. In the browser I can see an error where the common module is trying to export its functions.
> 
> The use of `require` in `common.js` can not be used together with `import` in `ProjectList.vue`. To make it work, use `import` and `export default` in `common.js` as shown below.

Eureka, thanks, just in time for v4.8.0!

nscuro:
> @valentijnscholten Should we go for lowercase badges in other places such as the Project view as well?

Ha, welp, I mixed up upper- and lowercase. Let's keep it uppercase for now so we're not messing with existing styling.

valentijnscholten:
> > @valentijnscholten Should we go for lowercase badges in other places such as the Project view as well?
> 
> Ha, welp, I mixed up upper- and lowercase. Let's keep it uppercase for now so we're not messing with existing styling.

OK, now I'm going with camelCase just to mess with you.

nscuro:
> OK, now I'm going with camelCase just to mess with you.

Why not make it [mocking-case](https://github.com/zakuradev/mocking-case)?

src/views/portfolio/projects/ProjectList.vue
```diff
@@ -179,6 +179,22 @@
               return `<a href="${url}">${xssFilters.inHTMLData(value)}</a>`;
             }
           },
+          {
+            title: this.$t('message.tags'),
+            field: "tags",
+            sortable: false,
+            visible: false,
+            formatter(value, row, index) {
+              let tag_string = row.tags.slice(0, 2)?.map(tag => common.formatProjectTagLabel(tag)).join(' ') || '';
```
nscuro, line 188: There needs to be an `undefined` / `null` check here, this line is currently failing when a project does not have any tags.

image

sephiroth-j:
> Should we go for lowercase badges in other places such as the Project view as well?

Switching to lowercase is #168.

src/views/portfolio/projects/ProjectList.vue
valentijnscholten: I am not seeing that error, but added a check anyway.

nscuro (approved):

PR #320, opened 2022-11-05
ACL: Make project names clickable

valentijnscholten:
When managing ACLs, the projects assigned to teams has a non-clickable name.
It thought it might be handy to make them clickable.
I did encode the url for XSS, but I am not sure this is needed when letting Vue take care of rendering the template.
(And it would be hard to embed a XSS payload in a uuid generated by the database)

2022-11-05 19_30_38-Dependency-Track - Administration


nscuro:
> I did encode the url for XSS, but I am not sure this is needed when letting Vue take care of rendering the template.

According to the Vue documentation, values included via attribute binding (e.g. `:href="this.href"`) are also encoded: https://vuejs.org/guide/best-practices/security.html#what-vue-does-to-protect-you

So you don't have to do it here.

nscuro (approved):

PR #328, opened 2022-11-14
Implement treegrid to display parent-child-relationship in project list

rbt-mm:
### Description
**Project list:**
Adds a treegrid to the project list table to display a hierarchical view, which shows the (already existing) multi-level parent-child-relationships between projects with lazy-loading in between levels.

The project list is hierarchical on default, but can be toggled to the old flat view.

Sorting and filtering by tags or classifiers works with the treegrid and the flat view, searching automatically switches to the flat view but preserves the state before the search and switches back to it when nothing is searched.

**Introduces new parent-child relationship logic:**
-   Prevents that inactive projects can be selected as parents
-   A project cannot be set to inactive if it has an active child
-   A parent project cannot select one of his children as a parent

**Introduces new logic to `ProjectCreateProjectModal`:**
Adds a new select input in the dialog for creating a project, which includes every active parent and creates a persistent relationship between them on selection.
Licenses and parents are now only retrieved when the dialogue is opened to reduce the amount of data which needs to be loaded for the project list.

**Introducing new logic to `ProjectDetailsModal`:**
Adds a new select input in the dialog for showing a projects details, which include every possible parent for the current project and creates a persistent relationship between them on selection.
Parents are now only retrieved when the dialogue is opened to reduce the amount of data which needs to be loaded for a project view.

hierarchical view active

hierarchical view inactive

flat view

create project

detail view

### Addressed Issue
[#84](https://github.com/DependencyTrack/dependency-track/issues/84)

### Additional Details
[Backend PR](https://github.com/DependencyTrack/dependency-track/pull/2152)

### Checklist


package.json
```diff
@@ -36,6 +36,7 @@
     "jquery": "3.6.0",
     "lodash-es": "4.17.21",
     "oidc-client": "1.11.5",
+    "jquery-treegrid": "0.3.0",
```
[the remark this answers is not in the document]
rkg-mm: @nscuro I can't see what exactly sonatype thinks to see here. From the DBs I checked I don't find a direct vuln in treegrid. Also it has only 4 dev-dependencies and 0 production dependencies according NPM. what is the issue here?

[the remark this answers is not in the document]
nscuro: Lift is showing vulns related to `qunit` and `bootstrap`, but nothing related to jQuery or even this specific `jquery-treegrid` component. The console is also showing some internal server errors for me, looks like a bug on their side.

src/views/portfolio/projects/ProjectList.vue
```diff
@@ -2,16 +2,18 @@
   <div class="animated fadeIn" v-permission="'VIEW_PORTFOLIO'">
     <portfolio-widget-row :fetch="true" />
     <div id="projectsToolbar" class="bs-table-custom-toolbar">
-      <b-button size="md" variant="outline-primary" v-b-modal.projectCreateProjectModal v-permission="PERMISSIONS.PORTFOLIO_MANAGEMENT">
+      <b-button size="md" variant="outline-primary" @click="initializeProjectCreateProjectModal" v-permission="PERMISSIONS.PORTFOLIO_MANAGEMENT">
         <span class="fa fa-plus"></span> {{ $t('message.create_project') }}
       </b-button>
       <c-switch style="margin-left:1rem; margin-right:.5rem" id="showInactiveProjects" color="primary" v-model="showInactiveProjects" label v-bind="labelIcon" /><span class="text-muted">{{ $t('message.show_inactive_projects') }}</span>
+      <c-switch style="margin-left:1rem; margin-right:.5rem" id="showFlatView" color="primary" v-model="showFlatView" label v-bind="labelIcon" :disabled="isSearching" v-b-tooltip.hover :title="$t('message.switch_view')" /><span class="text-muted">{{ $t('message.show_flat_view') }}</span>
```
nscuro, line 9: I think this switch should be toggled to `on` when searching, in addition to being disabled. 

It's confusing to have the view change to a flat list while "Show flat project view" is `off`.

src/views/portfolio/projects/ProjectList.vue
```diff
@@ -51,26 +59,79 @@
...
+              treeColumn: 0,
+              initialState: 'collapsed',
+            })
+          }
+          this.$refs.table.getData().forEach(project => {
+            if (project.children && !project.fetchedChildren && (this.showInactiveProjects || project.children.some(child => child.active))
+              && (!this.$route.query.classifier || project.children.some(child => child.classifier === this.$route.query.classifier))
+              && (!this.$route.query.tag || project.children.some(child => child.tag === this.$route.query.tag))) {
+              this.$refs.table.$table.find('tbody').find('tr.treegrid-' + project.id.toString()).addClass('treegrid-collapsed')
+              this.$refs.table.$table.find('tbody').find('tr.treegrid-' + project.id.toString()).treegrid('renderExpander')
```
nscuro, line 94:
```suggestion
              this.$refs.table.$table.find('tbody').find('tr.treegrid-' + project.uuid).addClass('treegrid-collapsed')
              this.$refs.table.$table.find('tbody').find('tr.treegrid-' + project.uuid).treegrid('renderExpander')
```

src/views/portfolio/projects/ProjectList.vue
```diff
@@ -51,26 +59,79 @@
...
+            }
+          })
+          this.$refs.table.getData().forEach(row => {
+            if (row.expanded) {
+              this.$refs.table.$table.find('tbody').find('tr.treegrid-' + row.id.toString()).treegrid('expand')
+            } else if (row.expanded === false) {
+              this.$refs.table.$table.find('tbody').find('tr.treegrid-' + row.id.toString()).treegrid('collapse')
```
nscuro, line 101:
```suggestion
              this.$refs.table.$table.find('tbody').find('tr.treegrid-' + row.uuid).treegrid('expand')
            } else if (row.expanded === false) {
              this.$refs.table.$table.find('tbody').find('tr.treegrid-' + row.uuid).treegrid('collapse')
```

src/views/portfolio/projects/ProjectList.vue
```diff
@@ -173,6 +234,9 @@
         ],
         data: [],
         options: {
+          idField: 'id',
+          parentIdField: 'pid',
```
nscuro, line 238: Did some testing and it seems like this does not have to be a numeric ID. We may just as well use the UUIDs, which IMO is preferable. The DT API does not expose technical IDs anywhere else, and I'd prefer if it stays that way.

```suggestion
          idField: 'uuid',
          parentIdField: 'parentUuid',
```

nscuro (changes requested):
This looks way better, thanks @rbt-mm!

Few suggestions, also the build is currently failing with:
```
npm ERR! `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm ERR! 
npm ERR! Missing: jquery-treegrid@0.3.0 from lock file
```

src/views/portfolio/projects/ProjectList.vue
rbt-mm: I initially tried to use UUIDs to display the child-parent-relationship, however doing so led to some strange behavior, which did not occur if a numeric ID is used.

I've just tried to use UUIDs again, and still get the same bug during testing.
If you look at the attached screenshot, some projects get duplicate entries when expanded and mimic the behavior of their identical counterpart. This only happens when using UUIDs, numerical IDs work just fine, but I cannot explain why or how it happens 😅 

**Behavior with UUIDs:**
image

**Behavior with numerical IDs:**
image

src/views/portfolio/projects/ProjectList.vue
rkg-mm: We intentionally left it at the current state, so we can switch back easier when search is cleared. But i think we can store the state somehow and revert to it differently instead

src/views/portfolio/projects/ProjectList.vue
nscuro: Ah I see, that's unfortunate of course. Alternatively, how about generating a numeric checksum from the UUID, similar to Java's `hashCode` or Murmur2 and use that instead? If the issue is about numbers vs. strings, then this could help.

rbt-mm: Using the `hashCode` of the UUIDs as values for `id` and `pid` seems to work well 👍
Would you still keep the names of the `JsonProperties` as `id` and `pid`?

nscuro: I think the hashing should be done on the client side, not in the API server. So the `id` and `pid` fields should be removed from the API response. After all this is a workaround for a limitation in the client.

src/views/portfolio/projects/ProjectCreateProjectModal.vue
```diff
@@ -166,25 +182,55 @@
...
+              if (this.project.parent && this.project.parent.uuid === project.uuid ) {
+                this.selectedParent = project.uuid;
+              }
+            }
+          }).catch((error) => {
+            this.$toastr.w(this.$t('condition.unsuccessful_action'));
+          }).finally(() => {
+            resolve()
+          });
+        })
```
nscuro, line 233: It just now dawned on me that this will either not load enough projects, or will be an immensely expensive operation in large portfolios.

The default `pageSize` for paginated APIs is 100. So if there are 105 projects, only the first 100 will be fetched here. You may get unlucky and the current parent is not among those 100 projects, causing the selected value to wrongly be empty.

If this limitation was not there, then we'd load too much data at once, and the dropdown would grow to an unmanageable size.

I'm wondering if maybe this should be a search field instead. The API server has a `/api/v1/search/project` endpoint that could be used:

https://github.com/DependencyTrack/dependency-track/blob/9c3290998e09c29ab4ffdb60cf1632ec4b72b07b/src/main/java/org/dependencytrack/resources/v1/SearchResource.java#L67-L83

Thoughts?

nscuro: Good question... After some quick googling, this looks promising: https://vue-multiselect.js.org/#sub-asynchronous-select

Not sure how well this blends in with the bootstrap-based UI though.

rbt-mm: I think that `vue-multiselect` would be a very good fit for this case and should blend in well with the already existing UI, if correctly adjusted.

However, I'm not quite sure about using the `/api/v1/search/project` endpoint for searching a parent project.
The endpoint uses `PROJECT_UUID`, `PROJECT_NAME`, `PROJECT_VERSION` and `PROJECT_DESCRIPTION` as search criteria (and can't be turned off, as far as I've seen it) for a project, but only `PROJECT_NAME` and `PROJECT_VERSION` are required in this dialogue, so the other criteria could match projects which wouldn't make sense for the user and create unnecessary results.

Another problem would be, that the endpoint only returns a projects UUID, name, version and description, but not its active status, which is essential for the dialogue because inactive projects cannot be selected as parents. The active status of a project could be added to the IndexConstants, but I don't know if this would cause any problems in other existing logic.

nscuro: I think you're right about the search endpoint. Even adding the `active` field would still cause more elements being returned by the API than the UI will actually display (filtering on client-side is kinda meh). And the current search endpoint will always perform a "contains" query, without users being able to specify which field to search in.

But you may be able to use the `/api/v1/project?searchText=<fieldInput>&excludeInactive=false` endpoint, which is also what the search field in the project list uses:

Screenshot 2022-11-27 at 22 41 39

src/views/portfolio/projects/ProjectCreateProjectModal.vue
```diff
@@ -166,30 +185,59 @@
...
+        if (project.version){
+          return project.name + " : " + project.version
+        } else {
+          return project.name
+        }
+      },
+      asyncFind: function (query) {
+        if (query){
+          this.isLoading = true
+          let url = `${this.$api.BASE_URL}/${this.$api.URL_PROJECT}?searchText=${query}&excludeInactive=false`
```
nscuro, line 231: Terribly sorry, I made a mistake when I initially shared the endpoint URL with you. 😅 It should be `excludeInactive=true` in your case, so that inactive projects are not queried and returned by the API. 

But I just tested it, and other than it currently returning inactive projects, it works great! Awesome job! 🙌

Issue #85, opened 2021-08-02
Dependency Graph does not follow transitive dependencies


exief:
### Current Behavior:
UI does not go through each transitive dependency in the dependency data to build out a complete dependency graph, resulting in a limited depth of dependencies shown in the UI.

### Proposed Behavior:
UI should follow each transitive dependency in the data, check if it has further dependencies, and render a complete dependency graph in the UI.

stevespringett:
@coderpatros its in [src/views/portfolio/projects/ProjectDependencyGraph](https://github.com/DependencyTrack/frontend/blob/master/src/views/portfolio/projects/ProjectDependencyGraph.vue)

claudioandreantonio:
Hi @stevespringett . Do you have an estimation when this might get fixed?
Thank you.

stevespringett:
@claudioandreantonio I'm not actively working on it and without a community PR, it will likely get pushed back to v4.6.

stevespringett:
Correct. I think the only thing necessary is to make calls to the server for all child dependencies whenever a parent dependency is expanded. That way, the plus arrow and all dependencies will be in place should the user request that information.

rstmm:
Unfortunately, the graph is a little useless anyway in this way in my opinion. 
The regular use case of the graph is: "Component X is vulnerable, I want to see where this dependency is coming from".
Even if the full graph would be printed, it will be hard to find the dependency (we regularly have 200-800 indirect dependencies!).

What would really be required is:
1. Printing the full graph
2. Having a button in the component list and component view like "Show in dependency-graph"
3. Focus the dependency-graph on that component, means: open the graph down to the correct component and highlight it 
4. Ideally additionally: grey out all components, which are not part of the parent-graph of this single component, so you got a highlighted component, a bright list of the upper tree leading there, and greyed out everything else.

Maybe that would require to change how the graph is displayed, make the elements much smaller and allow zooming, and moving left/right.

lsauer:
Some of my suggestions from #212 

**Current Behavior:**
- the tree has a cutoff at 3rd depth level. Or there is a bug or something I did not understand.
- double click on a node only expands
- the package name is not hyperlinked

**Proposed Behavior:**
- don't cutoff at any level and preload the entire tree
- double click on a node expands all subnodes
- the package name is hyperlinked and references the package entry page

For discussion:
Color the node-circle depending on the sum or avg of its child severities

rkg-mm:
[...]

PR #329, opened 2022-11-16
Show every depth level of dependency graph and improve readability

rbt-mm:
### Description
**Current Behavior:**
A project's dependency graph can only be expanded up to its third level, which makes it hardly usable and the size of the elements inside the graph makes it unnecessarily big and harder to read/navigate.
Also, the data for every node is requested at the same time, which can lead to huge traffic in projects with many dependencies.

**Proposed Behavior:**
The dependency graph of a project can be fully expanded to search for every dependency and only the necessary data for the expanding node is lazily loaded to reduce the amount of traffic.  Every path in the graph is also checked for circular references and prevents displaying them. 
The size of the buttons, the nodes, the font and other elements is reduced to improve the readability for bigger dependency graphs.

### Addressed Issue
#85 

### Additional Details

image

A PR will follow, that attaches a `Show in dependency graph` button to a component in different views, which redirects a user to the project's dependency graph and then expands the path to every occurrence of the selected component. 

### Checklist


nscuro (approved):
Awesome, thanks so much @rbt-mm! 🙌

PR #340, opened 2022-12-02
Respect project hierarchy in policy evaluation

rbt-mm:
### Description

Enables to set the option for a policy to also evaluate the children of the projects, to which the policy is limited to. 
This behavior can be toggled in the `Limit to`-dialogue of a policy or via the API.

### Addressed Issue

Backend Issue: [#2215](https://github.com/DependencyTrack/dependency-track/issues/2215)

### Additional Details

image

[Backend PR](https://github.com/DependencyTrack/dependency-track/pull/2223)

### Checklist


nscuro (commented):
Thanks for the PR @rbt-mm! 

The "Limit To" section is made invisible whenever the "Include Children" toggle is switched, which is a bit confusing. Can we prevent that from happening?

rbt-mm:
No problem @nscuro. 👍
Switching the "Include Children" toggle works correctly now.

nscuro (approved):

PR #346, opened 2022-12-06
Implement adding/deleting custom licenses via API

rbt-mm:
### Description

Enhancement to backend PR: [#2153](https://github.com/DependencyTrack/dependency-track/pull/2153)

Custom licenses can be added and deleted via the API, if a user has `SYSTEM_CONFIGURATION` permissions. Only custom licenses can be deleted and when a custom license is deleted, every policy condition, which has the license as the value, will also be deleted.

### Additional Details

To ensure that every policy condition, which has the deleted custom license assigned as the value, is also deleted, the backend method `deleteLicense()` of the original PR is enhanced: [Backend PR]()

image

image

image

image

### Checklist


src/views/portfolio/licenses/License.vue
```diff
@@ -56,6 +56,7 @@
         <div class="headerText formattedLicenseContent">{{ license.standardLicenseHeader }}</div>
       </b-tab>
     </b-tabs>
+    <b-button v-if="license.isCustomLicense" variant="outline-danger" @click="removeCustomLicense" v-permission="PERMISSIONS.SYSTEM_CONFIGURATION">{{ $t('message.delete_custom_license') }}</b-button>
```
nscuro, line 59: I think this should just say "Delete", and the "Custom License" part should be a checkbox like the *OSI Approved* and *FSF Libre* properties above.

nscuro (changes requested):

nscuro (approved):
Thanks @rbt-mm! 🙌

PR #359, opened 2022-12-12
Show version status information on dependency graph nodes

### LINKED ISSUE #358: Display version status on dependency graph nodes
### Current Behavior

The information about the version status of a component is currently not displayed in the dependency graph, forcing users to switch to a different project tab, if they want to know about the version status of a specific component.

### Proposed Behavior

It should be possible to optionally display the version status information for outdated components directly on the nodes of the dependency graph, to remove the need to switch to other tabs to check this information.

### Checklist

rbt-mm:
### Description

The information about the version status of a component is currently not displayed in the dependency graph, forcing users to switch to a different project tab, if they want to know about the version status.

This PR enhances the dependency graph to optionally display the version status information of a component on its corresponding node.

### Addressed Issue

Closes #358

### Additional Details

To display this information, the `/component/<componentUUID>` and the `/component/project/<projectUUID>/dependencyGraph/<componentUUID>` endpoints in the backend must be enhanced: [Backend PR](https://github.com/DependencyTrack/dependency-track/pull/2273)

**Note:** For the sake of better readability, the version status icon will only be shown for outdated components.

image

image

image

### Checklist


src/i18n/locales/en.json
```diff
@@ -509,7 +509,8 @@
     "show_in_dependency_graph": "Show in dependency graph",
     "show_complete_graph": "Show complete graph",
     "not_found_in_dependency_graph": "Dependency could not be found in dependency graph",
-    "reindex": "Rebuild index(es)"
+    "reindex": "Rebuild index(es)",
+    "show_update_information": "Show update information"
```
nscuro, line 513: > @rbt-mm For the sake of better readability, the version status icon will only be shown for outdated components.

Given this constraint, I think this should say "highlight outdated components" or so. "Show update information" is rather broad, we should be more specific here.

src/views/portfolio/projects/ProjectDependencyGraph.vue
```diff
@@ -293,7 +302,7 @@ export default {
     },
     getDependencyUrl(dependency) {
       if (dependency.objectType === "COMPONENT") {
-        return `${this.$api.BASE_URL}/${this.$api.URL_COMPONENT}/${dependency.uuid}`;
+        return `${this.$api.BASE_URL}/${this.$api.URL_COMPONENT}/${dependency.uuid}?includeRepositoryMetaData=true`;
```
nscuro, line 305: `includeRepositoryMetaData` is hardcoded to `true`. It should use `fetchRepositoryMetaData` instead.

nscuro (changes requested):
Thanks @rbt-mm! Two minor comments.

src/views/portfolio/projects/ProjectDependencyGraph.vue
rbt-mm: `includeRepositoryMetaData` is meant to be always `true` in this case, because it allows to seemingly switch between highlighting all outdated components or not, regardless of the state of the toggle when expanding a node and fetching its children.
If fetching the metadata was dependent on `fetchRepositoryMetaData`, then only the components which were fetched while the toggle was set to active would be highlighted as oudated components, which would be very confusing for the user.

`fetchRepositoryMetaData` is definitely a very bad name for this variable, because its only purpose is to check if the information for the outdated components should be rendered or not.
It has actually nothing to do with fetching repository metadata, therefore I would change the name of it to `highlightOutdatedComponents`, but still keep the `includeRepositoryMetaData` in the URL hardcoded to `true`.

nscuro: Oh ok, got it. Makes sense. 👌

Agreed that changing the name would be good.

nscuro (approved):

PR #378, opened 2022-12-30
minor: Enabled Sorting For Project Name in ComponentSearch.vue

kekkegenkai:
Signed-off-by: kekkegenkai <kekkegenkai@gmail.com>

### Description

The current "Components"-View does not allow users to sort the table by the project name.
Tis PR sets the "sortable" flag from false to true. 

The original issue (https://github.com/DependencyTrack/dependency-track/issues/353) was for the old frontend.
However, with the new frontend the issue seems to be by setting the flag.

### Addressed Issue
https://github.com/DependencyTrack/dependency-track/issues/353

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/github-templates/CONTRIBUTING.md#pull-requests)

### DIFF
--- a/src/views/portfolio/components/ComponentSearch.vue
+++ b/src/views/portfolio/components/ComponentSearch.vue
@@ -150,7 +150,7 @@
           {
             title: this.$t('message.project_name'),
             field: "project.name",
-            sortable: false,
+            sortable: true,
             formatter(value, row, index) {
               let url = xssFilters.uriInUnQuotedAttr("../projects/" + row.project.uuid);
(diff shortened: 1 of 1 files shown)


stevespringett:
This will not work the way you think it will, as the sort will only be applicable to the current paginated list of projects. The performance impact on the server could be extreme if this is set to true and the server had the query logic to sort by project. 

For this reason, sorting by project was disabled.

kekkegenkai:
Thanks @stevespringett for your feedback. I was not aware of that. Then this is the wrong PR.  I am closing this PR here.

PR #427, opened 2023-02-24
Add dualstack nginx listen

Meroje:
### Description

When testing over an ipv6-only stack I noticed the frontend would not be reachable, as nginx listens on ipv4 only by default. This change adds the second required directive.

### Checklist

- [/] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly

### DIFF
--- a/docker/etc/nginx/conf.d/default.conf
+++ b/docker/etc/nginx/conf.d/default.conf
@@ -1,5 +1,6 @@
 server {
   listen      8080;
+  listen      [::]:8080;
   server_name _;
 
   location / {


rissson:
We are also hitting this issue. This makes it work in all situations, ipv4-only, ipv6-only, and dualstack. Any chance this might get some attention?

nscuro (approved):
Thanks @Meroje!

nscuro:
@schmidi In what sense is it breaking? I was under the assumption that it should work for both IPv4 and IPv6.

schmidi:
> @schmidi In what sense is it breaking? I was under the assumption that it should work for both IPv4 and IPv6.

When I try to start the container on a server without IPv6 I get the following message:

`
dtrack-frontend_1   | /docker-entrypoint.sh: Configuration complete; ready for start up
dtrack-frontend_1   | 2023/10/23 09:34:01 [emerg] 1#1: socket() [::]:8080 failed (97: Address family not supported by protocol)
`

nscuro:
Found the issue as well as the solution: https://github.com/DependencyTrack/frontend/pull/621

Will backport this to 4.9.1. Thanks for raising the issue @schmidi!

schmidi:
@nscuro Thanks for the fast reply and the fix.

Meroje:
Well sorry about that, it never failed me but I guess I usually let ipv6 be un-routable rather than disabling it outright. Thanks for putting in in a better fix ❤️

PR #437, opened 2023-03-06
Add "Outdated only" and "Direct only" switch to only show outdated components

### LINKED ISSUE #311: Accessibility Issue with "Latest Version" Icon
### Current Behavior:
The Components Tab for projects displays a "Latest Version" icon as seen here:

image

The use of images in this way is an accessibility issue as the only way to distinguish between icons is by colour.  

### Steps to Reproduce:

Browse to any project and vew the latest version icons.

It is possible to enable the "Latest Version" column which displays actual version numbers...  but then one has to perform a comparison on every component to pick out which are actually out of date.  Much slower.

### Expected Behavior:
There is nothing wrong with colour, per se.  A way to improve accessibility would be to combine with **shape**.  There are a lot of excellent resources on how best to select shapes to indicate status, as shapes can also have cultural significance.

eg  square or tick (green) could be used to indicate "up to date".  Note also that this current status currently has an exclamation mark on it...  which can be a bit confusing to some as this symbol is often associated with a warning of some sort.

If shapes ARE used then I would hope that would allow the possibility of having the version status displayed in a separate column so that they can be sorted...  allowing everything that is not up to date to be grouped together for improved usability.  This latter logged as  #322 

### Environment:
  
  - Dependency-Track Version: 4.6.1
  - Client Browser:  Firefox 106
  - Client O/S: Windows 10

### LINKED ISSUE #436: New switches to only show outdated components and only show direct components in the project's components overview
### Current Behavior

The project's components overview always shows all components

### Proposed Behavior

Add a switch to only show outdated direct components
Add a switch to only show direct dependencies. One would want to update the direct dependencies when updates are available. Outdated transitive dependencies would only be updated when critical vulnerabilities exist and can be handles separately this way.

### Checklist

walterdeboer:
### Description

Added "Outdated only" switch and "Dircet only" switch in the project's components overview to only show outdated components and/or only direct dependencies

### Addressed Issue

Closes #436, fixes #311, addresses https://github.com/DependencyTrack/dependency-track/issues/208

### Additional Details
Also changed the icons do clearly distinquish outdated, unstable or recent components (#311).

Also reset the pageNumber to 1 in refreshTable (minor bug)

Needs https://github.com/DependencyTrack/dependency-track/pull/2568

image

image

### Checklist

- [x] This PR introduces new or alters existing behavior, ~and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly~


nscuro:
@walterdeboer Similar to #440, could you also resolve the merge conflicts in this PR please?

package.json
```diff
@@ -15,7 +15,8 @@
     "serve": "npm run prebuild && vue-cli-service serve",
     "prebuild": "node update-embedded-version.js",
     "build": "npm run prebuild && vue-cli-service build",
-    "lint": "vue-cli-service lint"
+    "lint": "vue-cli-service lint",
+    "sbom": "cyclonedx-npm --output-file sbom.json"
```
nscuro, line 19: Any reason why this was added? We already produce an SBOM with webpack.

nscuro (changes requested):
Looks good, just need clarification on the `sbom` command that was added.

package.json
walterdeboer: Nop, no reason at all. It's gone now :-)

nscuro (approved):

PR #502, opened 2023-05-23
fix login redirects

sephiroth-j:
### Description

Not having JWT was treated the same as not having permission to view the requested page.
Also, a small refactoring was done to make all possible logical paths clearer and the permission check was fixed as well.

### Addressed Issue

fixes #501 

### Additional Details

I also took the liberty of simplifying/modernizing `hasPermission` a little.

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#pull-requests)
- [ ] ~This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly~

PR #506, opened 2023-05-29
fix: no defined permission for dashboard

sephiroth-j:
### Description
dashboard was treated as a public route although it's not

### Addressed Issue
fixes regression from f16c314f4e0fde8c04d6b983fbff2f283b96c8b2

### Additional Details

Since #502, any route without the `meta.permission` property is treated as a public route. I did not check the dashboard route to see if it met this requirement. As a result, when calling the dashboard, there was no redirection to the login page if you were logged out, which caused consecutive errors on the dashboard page.

### Checklist

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -88,7 +88,8 @@ function configRoutes() {
           meta: {
             title: i18n.t('message.dashboard'),
             i18n: 'message.dashboard',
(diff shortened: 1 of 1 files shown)


nscuro (approved):
Good catch @sephiroth-j! Thanks for the PR.

PR #509, opened 2023-05-30
Fix affected projects tab not being updated when switching between vulnerability aliases

### LINKED ISSUE #481: Affected projects tab not updated when switching between aliases
### Current Behavior

When viewing a vulnerability, it shows the number of affected projects in the header of that tab. When switching to an alias the screen gets updated with the data from the alias, but the affected project tab does not get updated. The affected number of projects is/will probably be wrong in this case.

### Steps to Reproduce

1. Go to a vulnerability that has some affected projects and at least one alias, i.e. https://dt/vulnerabilities/NVD/CVE-2023-29197
2. Observe affected project count = X
3. Click on an alias
4. Observe affected project count is still X
5. Press F5
6. Observe affected project count is (usually) Y 

### Expected Behavior

The affected projects tab should be updated when switching between aliases.
I can see in the Bootstrap code this tab is "hardcodedly injected" and not tied to any bootstrap lifecycle or events.

4.8.0

### Browser

Google Chrome

### Browser Version

_No response_

### Operating System

Windows

### Checklist

- [X] I have checked the [existing issues](https://github.com/DependencyTrack/frontend/issues) for whether this defect was already reported

sephiroth-j:
### Description

The affected projects tab (and table) is not updated correctly when switching between aliases.

### Addressed Issue

fixes #481

### Additional Details

I'm leaving this as a draft for now because there is also a backend issue that should be fixed first (see https://github.com/DependencyTrack/frontend/issues/481#issuecomment-1568924285).

### Checklist

- [ ] DependencyTrack/dependency-track/issues/2794 is fixed

### DIFF
--- a/src/views/portfolio/vulnerabilities/AffectedProjects.vue
+++ b/src/views/portfolio/vulnerabilities/AffectedProjects.vue
@@ -152,6 +152,10 @@ export default {
       this.currentPage = 1;
       this.refreshTable();
     },
+    vulnId() {
+      // update url when vulnId changes, will trigger table refresh
+      this.$refs.table.refreshOptions({...this.options, url: this.apiUrl()});
+    },
   },
 };
 </script>


sephiroth-j:
Hello @valentijnscholten, would you please check if this actually fixes the problem you described in #481?

valentijnscholten:
The response in affected projects from the backend will be different for an alias, that's the reason the tab must be refreshed.
If the responses were the same, a refresh of the tab would not be needed. I will test if the fix works.

valentijnscholten:
@sephiroth-j Sorry for the late reply, needed some time to get DT going again locally and be able to build the PR.
But it doesn't seem to work. I think I have tried this solution as well, but the `props` aren't changing when the user switches to the alias.

valentijnscholten:
Oh sorry, there were some errors running `npm run serve`. Just reapplied your changes onto master and it works!
Thanks. Can you rebase?

sephiroth-j:
Hi @valentijnscholten, I have rebased the bugfix branch to 4.12.x and updated the target branch of the PR accordingly.

valentijnscholten (approved):
thanks, it still works :-D

valentijnscholten:
@sephiroth-j Can you sign your commits?

sephiroth-j:
@valentijnscholten my commit is signed. I think you just need to re-run the DCO check.

nscuro:
I have manually passed the DCO check since @sephiroth-j's commit is indeed signed off.

PR #511, opened 2023-06-01
Create static/config.json from template file + ENV if not exist

### LINKED ISSUE #60: improve entrypoint
### Current Behavior:

https://github.com/DependencyTrack/frontend/blob/master/docker/docker-entrypoint.sh#L6
checks for a mount, this does work well with configmaps in kubernetes. Configmaps is seen as a good practice.

### Proposed Behavior:
could it instead simply check that the file exists, and if it does, ignore it. Also note that in case of configmaps it is not a ordinary file, but rather a symlink:
```
/app/static # ls -l conf*
lrwxrwxrwx    1 root     root            18 Mar  2 19:56 config.json -> ..data/config.json
/app/static # ls -l /app/static/config.json 
lrwxrwxrwx    1 root     root            18 Mar  2 19:56 /app/static/config.json -> ..data/config.json
/app/static # ls -l /app/static/..data
lrwxrwxrwx    1 root     root            31 Mar  2 19:56 /app/static/..data -> ..2021_03_02_19_56_07.520397086
/app/static # 
```

muellerst-hg:
Fixes #60

### Description

`static/config.json` can be set either from ENV variables or by mounting the file using a volume. docker-entrypoint.sh takes care of that. Current behavior was that `config.json` is used as is if the file is mounted , otherwise the file is populated with ENV variables.

In some container environments, `mount` command is not feasible (e.g. kubernetes 1.27.2 / containerd 1.7.2) and entrypoint.sh fails.

Currently, `static/config.json` is build into the image. However it just serves as a template, because it requires at least API_BASE_URL to be set. So instead of building an empty `static/config.json` into the image, rather build the image with `config.json.tmpl` Then check in docker-entrypoint if config.json file exists. If yes, then just leave it as it is. If not, create the file from config.json.tmpl and populate with ENV variables.

### Addressed Issue

fixes #60

### Checklist


nscuro:
> To support container environments where mount command is not feasible

@muellerst-hg Under what conditions exactly will `mount` not work? I tried reproducing this in a related issue (https://github.com/DependencyTrack/frontend/issues/513#issuecomment-1587809550), but failed to do so locally. Any idea what I can do to reproduce this problem?

muellerst-hg:
@nscuro
kubernetes 1.27.2
containerd 1.7.2
runc 1.1.7

mount works well with docker 24.0.2

nscuro:
Thanks @muellerst-hg, I was able to reproduce with

```shell
minikube start --kubernetes-version v1.27.2 --container-runtime containerd
```

So it's a Containerd limitation.

We now have to approaches that aim to fix this:

* Your PR (#511): Checks if `config.json` is writeable to determine if it is mounted
* @Breee's PR (#514): Checks `/proc/mounts` directly instead of using `mount`

What would be the better approach? One thing that comes to mind for the `touch` way, is that folks using Docker Compose more often than not do not specify `:ro` for their volume mounts. In those cases the file will be mounted, but still writeable.

muellerst-hg:
Good catch regarding `:rw` volume mounts.

I would prefer a different approach:
Currently the image is build with `config.json`. However the file just serves as a template, because it requires at least `API_BASE_URL` to be set - either using a volume mount or by overriding it with `docker-entrypoint.`

Build the image with `config.json.tmpl` instead of `config.json`. Check in `docker-entrypoint` if `config.json` file exists. If yes, then just leave it as it is. If not, create the file from ENV

nscuro:
Sounds like a sane approach to me. @Breee, WDYT?

Breee:
> Sounds like a sane approach to me. @Breee, WDYT?

Yeah I agree. In addition I think that the frontend should not be broken, if there is no config, i.e. be able to handle an empty API_BASE_URL and inform that it is not configured.

muellerst-hg:
@nscuro updated the PR branch, PR title the description accordingly. Happy merge!

public/static/config.tmpl.json
nscuro, line 1: We need a way to do this without moving the file per default.

This move would break local development workflows, and require an additional manual step for folks who deploy the frontend without containers. Perhaps we can perform the move in `Dockerfile` instead?

muellerst-hg: I see. I'll come up with an update, but it may take some days

nscuro: I made the desired changes and raised #622. I included you as co-author as I only modified your PR. Thanks again for working on this.

muellerst-hg: Thanks for taking over .

Issue #513, opened 2023-06-07
mount  in entrypoint does not work in all environment


Breee:
### Current Behavior

Given the following kubernetes deployment: 

```
metadata:
  name: dependencytrack-dev-frontend
  namespace: dependencytrack-dev
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: dependencytrack
      component: frontend
      stage: dev
  strategy:
    type: Recreate
  template:
    metadata:
      annotations:
      labels:
        app: dependencytrack
        component: frontend
        stage: dev
    spec:
      containers:
      - image: dependencytrack/frontend:4.8.1
        imagePullPolicy: Always
        name: dependencytrack-dev-frontend
        ports:
        - containerPort: 8080
          name: http
          protocol: TCP
        securityContext:
          readOnlyRootFilesystem: true
          runAsGroup: 0
          runAsUser: 101
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /tmp
          name: tmp
        - mountPath: /etc/localtime
          name: tz-config
        - mountPath: /opt/owasp/dependency-track-frontend/static/config.json
          name: static-config
          subPath: config.json
      imagePullSecrets:
      - name: dockerhub
      restartPolicy: Always
      volumes:
      - emptyDir: {}
        name: tmp
      - hostPath:
          path: /usr/share/zoneinfo/UTC
          type: ""
        name: tz-config
      - configMap:
          defaultMode: 420
          items:
          - key: config.json
            path: config.json
          name: dependencytrack-dev-static-config
        name: static-config
```
and the following configmap: 

```
apiVersion: v1
data:
  config.json: |-
    {
      "OIDC_ISSUER": "xxxxxxxxxxx",
      "OIDC_CLIENT_ID": "xxxxxxxxxxxxxxxxx",
      "OIDC_SCOPE": "openid profile email",
      "OIDC_FLOW": "",
      "OIDC_LOGIN_BUTTON_TEXT": "Login with Azure AD",
      "API_BASE_URL": "https://api.dependencytrack-dev.xxxxxxxxxxx"
    }
kind: ConfigMap
```

I experience the following logs: 

```
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
[...]

PR #514, opened 2023-06-07
Check /proc/mounts instead of mount

Breee:
### Description

We could use `cat /proc/mounts` as workaround  in case that the `mount` command does not work in some environments

### Addressed Issue

#513 

### Checklist


Breee:
closing in favour of https://github.com/DependencyTrack/frontend/pull/511

Issue #500, opened 2023-05-17
Export functionality to download metrics from components and vulnerabilities


nvcastelli:
### Current Behavior

The data we have in Dependency Track is used to share component, vulnerability and other information to our stakeholders. Since we need a way to transfer this information to other users we have the use case for us to be able to download some of this information in a readable way and share it with our colleagues.

### Proposed Behavior

If we were able to implement a way to download information to a spreadsheet from certain UIs in the application that would satisfy our use case. I personally implemented a feature like this and am hopeful to include it into dependency track itself, pending input from the team. Please see an image of how it looks below and note the download button above the table in the UI.
![Screen Shot 2023-05-17 at 3 15 46 PM](https://github.com/DependencyTrack/frontend/assets/3475905/3afd7ba1-6007-4166-afae-452c6e0dd872)

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/frontend/issues) for whether this enhancement was already requested

stevespringett:
Thanks for the suggestion. I would prefer to use the Export plugin in Bootstrap Table. It would save screen real estate and supports multiple formats.

Demo: https://examples.bootstrap-table.com/#extensions/export.html
API Docs: https://bootstrap-table.com/docs/extensions/export/

nvcastelli:
Hey @stevespringett , thank you for your comment, my apologies for taking this long to get back with you but I've been working on implementing the export plugin into my local instance of dependency track. I wanted to get back with some of my findings and have a dialogue with you on what features this enhancement would need to be included into the application.

One of the first things I've found with the export plugin is that it only exports the data that is shown on the table. For example, if you are looking at component with 100 items and have the table view amount set to 50, the export plugin only downloads the 50 items that are shown on the table. We were hoping that we'd be able to pull all of the data for a given component instead of just what we can see in the table.

Secondly, how the plugin is shown in the Dependency Track UI is causing some issues that I've opened a [ticket](https://github.com/hhurz/tableExport.jquery.plugin/issues/366) against the repo that maintains the plugin. l'm not sure if this is something on my end that just needs configuration or if it's something the export plugin will need to handle. I'm waiting for their response now.

Please let me know if there's any clarifications that I can make, thanks.

nvcastelli:
[...]

PR #526, opened 2023-06-14
Export table

nvcastelli:
### Description
This changes utilizes the bootstrap-table export plugin to pull table data from tables within the Dependency Track application. In places where it's been enabled on a bootstrap table users will be able to select a dropdown to download table data based on the file type they select.

### Addressed Issue
This PR is the first step to fulfilling the following enhancement #500. Pending approval this should resolve enhancement #500

### Additional Details
We had brought up this enhancement and Steve suggested using the bootstrap-table export plugin. Currently, this change only adds the ability to download on the component table and only in json and txt formats. I am looking into adding the ability to download as an excel but there are more dependancies needed.

For this change to work some new dependencies need to be included:
-  js-xlsx
- tableexport.jquery.plugin
- file-saver

I'm happy to add any suggestions.

### Checklist


nscuro:
Hi @nvcastelli, thanks for the PR!

> For this change to work some new dependencies need to be included: 
>   *  js-xlsx
>    * tableexport.jquery.plugin
>    * file-saver

Could you please add those dependencies to `package.json` via `npm install`, so this change can be tested?

nscuro:
@nvcastelli I just gave this a spin, but it seems like despite usage of `exportDataType: 'all'`, the export still only contains only up to 100 components.

The issue seems to be that, in order to fetch *all* items, the plugin requests the REST endpoint without `pageNumber` / `pageSize` query parameters, e.g.:

```
http://localhost:8081/api/v1/component/project/f803526f-fe36-4485-ae0b-e7fe463caf18?onlyOutdated=false&onlyDirect=false&searchText=&sortName=name&sortOrder=asc
```

Whereas in the paginated view it would do:

```
http://localhost:8081/api/v1/component/project/f803526f-fe36-4485-ae0b-e7fe463caf18?onlyOutdated=false&onlyDirect=false&searchText=&sortName=name&sortOrder=asc&pageSize=25&pageNumber=1
```

For paginated REST endpoints, DT will fall back to `pageNumber=1` and `pageSize=100` when not set explicitly:

https://github.com/stevespringett/Alpine/blob/cd6aa7ed05376935ab32bc43819eba0e3a525b7f/alpine-infra/src/main/java/alpine/persistence/Pagination.java#L58-L65

If the intention is to truly fetch *all* components, we'll need a way to deal with DT's default pagination behavior here.

nvcastelli:
Hey @nscuro , thank you for looking into it. My understanding is that the export-table functionality sets the pagesize to the amount of items in the list, then pulls from them that way. But if you're saying it only does that up to a maximum 100 that would be a problem. If we want to keep using these plugins we'd have to look into how DT does the pagination like you mentioned. Another option may be to just grab the data that's returned from the request to fill the table and put it into a file. Not sure if there's a good plugin for this, maybe something we'd have to make from scratch. Let me know your thoughts, and thanks again!

nscuro:
> My understanding is that the export-table functionality sets the pagesize to the amount of items in the list, then pulls from them that way.

In that case I'd have expected it to set `pageNumber=1&pageSize=N` for a project with N components, but that was not happening when I tested it. 

> If we want to keep using these plugins we'd have to look into how DT does the pagination like you mentioned.

Maybe there's a hook or callback we can use to manually populate the request parameters?

> Another option may be to just grab the data that's returned from the request to fill the table and put it into a file.

Yeah, at that point it may be easier to do it in the API server altogether. I personally do not have strong feelings about the bootstrap plugins, but I understand Steve would appreciate us making it work somehow using this plugin.

leec94:
hi @stevespringett looks like the recommended plugin isn't working as we'd like. there's some discussion around researching other directions to implement this, would you be open to us researching other plugins or getting the data from the backend instead?

Issue #562, opened 2023-07-31
Possibility to provide searchText to ProjectList via url query param


heubeck:
### Current Behavior

Searching the project list is only possible by using the search field of the project table.

### Proposed Behavior

It would be awesome, if the search text could provided via url query param as an entry point to a filtered ProjectList page.

Makes sence for similar list views: vulns, policies and licenses as well.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/frontend/issues) for whether this enhancement was already requested

PR #563, opened 2023-07-31
Issue #562: Propagate searchText query to List views

heubeck:
resolves https://github.com/DependencyTrack/frontend/issues/562


nscuro (approved):
Great enhancement, thanks @heubeck! 🤘

PR #619, opened 2023-10-21
Keep focus when editing Alerts (fixes #421)

### LINKED ISSUE #421: Alerts: Focus Lost when Editing Field
### Issue Type:
  - [X] defect report
  - [ ] enhancement request
  
### Current Behavior:
* Within DT v3.4.0 UI, navigate Administration -> Alerts -> Create Alert
* Create an alert with publisher "Email"
* Now open up your new alert and try to edit Destination.
* When you click in the "Destination" field, the border colour changes from yellow to blue, indicating that this field has focus.  This colour change is correct.
* If you start typing immediately and continiously,  then there are no problems...  you can enter the destination.
* The defect occurs should you stop typing (ie, looking away to double-check that what is being entered is correct):  after about 1 second, focus is lost and the field border changes back to yellow.

### Steps to Reproduce (if defect):
See above

### Expected Behavior:
Focus should not change automatically.  The user should be allowed to enter their data into the field.
Perhaps this control can be provided by adding a "submit" button to the screen?  This would allow the user to double-check everything that they have changed before committing anything.

### Environment:
  
  - Dependency-Track Version: 3.4.0
  - Distribution: [Executable WAR]
  - Database Server: [PostgreSQL]
  - Browser: Firefox 66.0.2

AbdelHajou:
### Description

When editing an Alert, the form lost focus after every 750ms. This worsened the user experience because the user had to keep re-selecting the form field to make changes. Instead of auto-saving after every 750ms, the form now has a Submit button and will no longer lose focus unintentionally.

### Addressed Issue
Fixes #421 

### Additional Details
afbeelding

### Checklist

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -653,6 +653,7 @@
     "notification_level": "Notification level",
     "destination": "Destination",
     "delete_alert": "Delete Alert",
+    "submit": "Submit",
     "limit_to": "Limit To",
     "limit_to_projects": "Limit to projects",
     "include_children": "Include children of projects",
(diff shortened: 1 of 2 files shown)


nscuro (approved):
Awesome, thanks @AbdelHajou!

PR #573, opened 2023-08-08
Add "Show in Dependency-Graph" Button in "Affected Projects" List

rbt-mm:
### Description

This PR adds the `Show in Dependency-Graph` button to the every project in the `Affected Projects` tab of a vulnerability, but only if the affected project has a dependency graph.
Clicking the button redirects the user to the projects dependency graph and highlights the affected component.

Clicking on the project link in a vulnerabilities `Affected Projects` tab opens the project's `Audit Vulnerabilities` tab, showing only the expanded finding.

### Addressed Issue

#533

### Additional Details

Backend PR: DependencyTrack/dependency-track#2942

image

\
The redirect to a project's `Audit Vulnerabilities` tab works by opening a specific URL and searching for the matrix of a finding.
By adding a hidden `Matrix` column to the table, the wanted finding can be filtered and expanded by searching for it's matrix.
- URL Format: `.../projects/<projectUUID>/findings/<affectedComponentUUID>/<vulnerabilityUUID>`
- Matrix Format: `<projectUUID>:<affectedComponentUUID>:<vulnerabilityUUID>`

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#pull-requests)
~- [ ] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly~

Issue #533, opened 2023-07-04
Usability Improvement: Add "Show in Dependency-Graph" Button in "Affected Projects" List


rkg-mm:
### Current Behavior

Feedback from our project users:
The current workflow is:

1. Get email of new vulnerability
2. Open "Vulnerability link" from mail
3. Read vulnerability details
4. Question yourself where this dependency is coming from
5. Switch to "Affected Projects" tab and select a project
6. Now you are in the Dashboard of the project
7. Switch to Vulnerabilities tab [Alternative: Components tab]
8. Find the vulnerability in the list of vulnerabilities [Alternative: Find affected component] - if you remember the details you need to identify it, otherwise go back to step 3...
9. Click "Show in Dependency-Graph" - now you see where the vulnerability is coming from

### Proposed Behavior

- Add a "Show in Dependency-Graph" directly in the list of affected projects tab in the vulnerability information
- Optional addition: When clicking the link of an affected project in the "affected projects"-tab, open the vulnerability tab of that project instead of the dashboard, ideally expand the vulnerability immediately and scroll to it, so user can directly start analysis decisions etc.

Expected Workflow:
1. Get email of new vulnerability
2. Open "Vulnerability link" from mail
3. Read vulnerability details
4. Question yourself where this dependency is coming from
5. Switch to "Affected Projects" tab and click **"Show in Dependency-Graph" Icon** from the project
6. Now you see where the vulnerability is coming from

Very small change but would increase user acceptance in projects a lot from what I hear, as this workflow is the most used one and feels unnecessary complicated with some frustration on the path

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/frontend/issues) for whether this enhancement was already requested

nscuro:
Great suggestion!

nscuro:
Implemented in #667. Thanks @rkg-mm!

PR #667, opened 2023-12-09
Add "Show in Dependency-Graph" Button in "Affected Projects" List [improved version]

rkg-mm:
### Description

This PR adds the Show in Dependency-Graph button to the every project in the Affected Projects tab of a vulnerability, but only if the affected project has a dependency graph.
Clicking the button redirects the user to the projects dependency graph and highlights the affected component.

Clicking on the project link in a vulnerabilities Affected Projects tab opens the project's Audit Vulnerabilities tab, showing only the expanded finding.

**Improved version, replaces** https://github.com/DependencyTrack/frontend/pull/573 :

- The button now supports opening all affected components if vulnerability affects more than one component. The backend is adapted to return all affected components.
- Additionally, refactored and simplified some code
- Additionally fixed a bug, which prevented outdated versions to be displayed in graph when opening it for searched components

### Addressed Issue
https://github.com/DependencyTrack/frontend/issues/533

### Additional Details

Backend PR: https://github.com/DependencyTrack/dependency-track/pull/3285

257805393-c33d4636-b330-4510-a4e1-4ee63b770d92

The redirect to a project's Audit Vulnerabilities tab works by opening a specific URL and searching for the matrix of a finding.
By adding a hidden Matrix column to the table, the wanted finding can be filtered and expanded by searching for it's matrix.

- URL Format: `.../projects/<projectUUID>/findings/<affectedComponentUUID>/<vulnerabilityUUID>`
- Matrix Format: `<projectUUID>:<affectedComponentUUID>:<vulnerabilityUUID>`

### Checklist


src/views/portfolio/projects/ProjectDependencyGraph.vue
```diff
@@ -64,78 +64,72 @@ export default {
...
+      if(!this.$route.params.componentUuids) {
+        data.children = this.transformDependenciesToOrgTree(JSON.parse(this.project.directDependencies), true, {gatheredKeys: []}, this.project.uuid, "PROJECT");
+        this.data = data;
+        return;
+      }
+
+      // tree with component search active
+      this.createSearchedComponentLookupTable(this.$route.params.componentUuids);
+      this.loading = true
+      let url = `${this.$api.BASE_URL}/${this.$api.URL_COMPONENT}/project/${this.project.uuid}/dependencyGraph/${this.$route.params.componentUuids}`
```
nscuro, line 101: It seems like `componentUuids` must be URL encoded. In my testing I am receiving 400s from the API when more than one component is requested. The 400 is returned even before the request reaches the `ComponentResource#getDependencyGraphForComponent` method.

For reference, here is an example URL:

```
http://localhost:8081/api/v1/component/project/d8e78c10-cf81-4733-a206-a0abc296c39a/dependencyGraph/dd65ef43-8c21-4ce9-9ff6-e04cec141eb5|fa3d689e-39af-46f8-b330-294b2da669dd
```

When I encode the pipe (`|`) as `%7C`, I am getting through to the API resource:

```suggestion
      let url = `${this.$api.BASE_URL}/${this.$api.URL_COMPONENT}/project/${this.project.uuid}/dependencyGraph/${encodeURIComponent(this.$route.params.componentUuids)}`
```

src/views/portfolio/vulnerabilities/AffectedProjects.vue
```diff
@@ -25,9 +26,20 @@
             title: this.$t('message.name'),
             field: "name",
             sortable: true,
-            formatter(value, row, index) {
-              let url = xssFilters.uriInUnQuotedAttr("../../../projects/" + row.uuid);
-              return `<a href="${url}">${xssFilters.inHTMLData(value)}</a>`;
+            formatter: (value, row, index) => {
+              const url = this.$router.resolve({name: 'Project Finding Lookup',
+                  params: {'uuid': row.uuid, affectedComponent: row.affectedComponentUuids[0],
+                      vulnerability:this.vulnerability}}).href;
```
nscuro, line 32: This is a seriously cool feature!

However, it suffers from the same limitation as the link to the dependency graph did before: If more than one component is vulnerable, the user is only taken to a single occurrence of the finding, which can be confusing.

Perhaps this button should only include the vulnerability UUID, and not the other parts of the matrix? This way, all occurrences of the vulnerability would be shown.

src/views/portfolio/projects/ProjectDependencyGraph.vue
rkg-mm: Interesting, in my case it seems to work without encoding because Chrome does encode it automatically. I end up at an encoded URL without doing it in the code. However, I added this now.

src/views/portfolio/vulnerabilities/AffectedProjects.vue
rkg-mm: I removed the limitation as suggested. Is it ok if only the first entry opens?

src/views/portfolio/projects/ProjectDependencyGraph.vue
nscuro: Ha, I was low-key expecting the browser to play a role here. I'm using Firefox. So annoying that we still have differing behavior in such basic functionalities across browsers...

src/views/portfolio/vulnerabilities/AffectedProjects.vue
nscuro: Yeah that should be fine.

rkg-mm:
@nscuro  Could you consider removing the code analysis rule triggering on things like
`formatter(value, row, index) {`
with
`'index' is defined but never used. `

I fixed the one in this PR but I see this failing other PRs, too. and I find it better you define all given parameters even if you do not use them right now, cause it makes it clear that they are available when you work on this part later.

nscuro:
@rkg-mm Codacy appears to be using ESLint behind the scenes, so you can just add `// eslint-disable-line no-unused-vars` to the offending line.

But yeah I think Codacy in general needs some tweaking to be a better fit for the project in general.

nscuro (approved):

PR #682, opened 2023-12-19
Use VUE_APP_SERVER_URL

mykter:
### Description

Make use of the VUE_APP_SERVER_URL from .env.development in the vue config.

### Addressed Issue

Fixes #681

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#pull-requests)
- [ ] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly

PR #683, opened 2023-12-19
Include component count in projects list

### LINKED ISSUE #630: Project Screen: Add Component Count Column
### Current Behavior

The project screen has a useful (sortable) column that displays project risk score.  However,  there is no way to tell at a glance which projects have high risk scores and a huge component count or low risk scores and low component count.

In my portfolio I have:

* Projects with 0 risk score where several have 0 components (which is something warranting investigation in and of itself)  and some have a few score components.
* Projects with similar high risk score where one or two have a low component count.  Which is which?
* Microservice projects that have almost identical components (and thus component counts).  But not always.... some have many extra components because test dependencies were not scoped correctly.

### Proposed Behavior

Add a sortable (and optional?) column to the projects page that displays the component count for each project.   This should help with all the above use cases.

### Checklist

mykter:
### Description

Include project component count on the projects listing page (hidden by default).

### Addressed Issue

Fixes #630

### Additional Details

The other metrics fields have specific code to handle the case where the metrics aren't provided; I don't think that is necessary in this case where just a number is being rendered, not a progress bar. Example with two projects, one with metrics and one without:

image

(The diff accidentally builds on #682)

### Checklist

### DIFF
--- a/src/views/portfolio/projects/ProjectList.vue
+++ b/src/views/portfolio/projects/ProjectList.vue
@@ -245,6 +245,12 @@ import ProjectCreateProjectModal from "./ProjectCreateProjectModal";
             sortable: true
           },
           {
+            title: this.$t('message.components'),
+            field: "metrics.components",
+            sortable: true,
+            visible: false
+          },
+          {
             title: this.$t('message.policy_violations'),
             field: "metrics",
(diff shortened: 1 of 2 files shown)


nscuro (approved):
Thanks yet again!

PR #708, opened 2024-01-28
fix(common): calculate percentage to tenth decimal place

setchy:
### Description

Refactor `calcProgressPercent` to found to the tenth decimal place, as per the functions comment/description.

```
/**
 * Given a total number of something and a lower number that is completed,
 * function will return a percentage rounded to the tenth decimal place.
 */
 ```

Also changed the order of the if statements to better flow.

### Addressed Issue

Previous logic was rounding to the nearest integer.

Previous logic 

### Checklist


nscuro (approved):
Thanks @setchy, appreciate all the tweaking!

PR #658, opened 2023-12-02
Introduce "collection" projects for better usage of hierarchical view

rkg-mm:
### Description
This change introduces UI logic for "collection projects". Those are basically projects used as parent for other projects that shall not hold any own component or vulnerability data, but instead get calculated from child projects using different configurable aggregation logics.

* Project Details/create dialog allow selection of different collection logics (all direct children, direct children with custom tag, direct children which are latest version)
* Collection projects are visually marked in project list and get shown aggregated metrics. The used calculation logic is displayed on mouseover of calc icon
* Project page hides all unrelevant tabs, instead shows a new tab with all child projects. For this the project list is made reusable
* Project metric charts display a blue vertical line when projects collection logic changes (includes entry in tooltip)
* Allow NONE as project classifier and
** Add Frontend support for other missing classifiers which are supported by backend already since 2023
** Move available classifiers into mixin to reduce code duplication & ensure classifier dropdown list is sorted identical in all views
* Additionally:
** Fixes some routing bugs

Required Backend PR: https://github.com/DependencyTrack/dependency-track/pull/3258

### Addressed Issue
https://github.com/DependencyTrack/dependency-track/issues/2041
https://github.com/DependencyTrack/dependency-track/issues/2410
#641 (routing Bug)
Fixes #1025 

### Additional Details

Hint: Screenshots show outdated "HighSemver" functionality, which was exchanged.

image

image

image

image

### Checklist
- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#pull-requests)
- [x] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly

PR #727, opened 2024-02-03
Fix table column visibility preferences triggering redundant requests

nscuro:
### Description

Fixes table column visibility preferences triggering redundant requests.

### Addressed Issue

N/A

### Additional Details

When applying user preferences for column visibility, fields of the bootstrap table were accessed directly. Apparently the bootstrap table has watches on those fields, which caused table contents to be loaded at least twice, instead of once.

This does not happen when using the "official" `table.showColumn` / `table.hideColumn` methods. Switching to those no longer causes duplicate requests.

### Checklist

- ~This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly~

### DIFF
--- a/src/shared/utils.js
+++ b/src/shared/utils.js
@@ -72,8 +72,24 @@ export function getContextPath() {
 }
 
 export function loadUserPreferencesForBootstrapTable(_this, id, columns) {
+  const table = _this.$refs.table;
+  if (!table) {
+    console.error("No table defined in the calling component; Can't apply user preferences");
+    return;
+  }
+
   columns.forEach((column) => {
-    _this.$set(column, "visible", (localStorage && localStorage.getItem(id + "Show" + common.capitalize(column.field)) !== null) ? (localStorage.getItem(id + "Show" + common.capitalize(column.field)) === "true") : column.visible);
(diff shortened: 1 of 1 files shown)


rkg-mm:
@nscuro does this fix the same as I fixed in #658?

nscuro:
Ah, sorry it didn't occur to me that you already identified and fixed this particular issue! 

Looks like our solutions are different but achieve the same thing. Reading the [Defer URL docs](https://bootstrap-table.com/docs/extensions/defer-url/) it looks like the more sustainable approach, so may use that instead of what I hacked together. 😆

rkg-mm:
Yes I came across this when implementing the other stuff and fixed it directly since it had an effect on what I did there :D maybe should have documented it better :D Yes I think that would be the better approach

nscuro:
Hmmm I just tested and `deferUrl` is not universally usable. For example in the project list, it causes the list to always be empty until a project is added or the filter toggles are used. Using the table's refresh button does nothing.

rkg-mm:
Hmm ok I didn't notice any issues :(

Issue #765, opened 2024-03-07
Aliases column is not displaying information in Vulnerabilities section


LaVibeX:
### Current Behavior

When a vulnerability has an alias, the column displays no information. This issue arises from a problem in the frontend code, specifically in line <73:VulnerabilityList.vue> const aliases = common.resolveVulnAliases(value); where the function expects both the vulnerability source and the vulnerability alias, but only receives one parameter.

![image](https://github.com/DependencyTrack/frontend/assets/52439101/de3202da-54e2-44d7-8230-ef76c4b31e3a)

### Steps to Reproduce

1. Go to the Vulnerabilities Section
2. Activate Alias
![image](https://github.com/DependencyTrack/frontend/assets/52439101/3ff1ce11-c31e-4faf-890d-67eecc9e0941)

3. If a vulnerability has any aliases, the Aliases column does not display them.

### Expected Behavior

Aliases Column displaying the alias assigned to that vulnerability in Vulnerabilities Section. 
![image](https://github.com/DependencyTrack/frontend/assets/52439101/6d76e289-3124-480f-be51-bac8094dd13e)

### Dependency-Track Frontend Version

4.7.x

### Browser

Google Chrome

### Browser Version

Firefox

### Operating System

Windows

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/frontend/issues) for whether this defect was already reported

nscuro:
Closed via https://github.com/DependencyTrack/frontend/pull/766

PR #766, opened 2024-03-07
Aliases column is not displaying information in Vulnerabilities section

LaVibeX:
### Description
Add Neccesary parameter to resolveVulnAliases function
common.resolveVulnALiases needs the vulnerability source and the aliases to deduplicate the list on the Aliases column,

The only parameter added to the function was "value" which do not include neither the source of the current vulnerability or their aliases.

Thats why I am adding row.source and row.aliases to the function.

I discovered that the 'return' statement was located within the condition if(typeof value !== 'undefined'), This created a problem because the function would not return anything for null values. To fix this, I moved the 'return' outside of the condition.

### Addressed Issue
#765 

### Checklist

### DIFF
--- a/src/views/portfolio/vulnerabilities/VulnerabilityList.vue
+++ b/src/views/portfolio/vulnerabilities/VulnerabilityList.vue
@@ -87,9 +87,9 @@ export default {
           field: 'aliases',
           visible: false,
           formatter(value, row, index) {
-            let label = '';
             if (typeof value !== 'undefined') {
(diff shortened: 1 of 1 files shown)


nscuro:
Thanks for the PR @LaVibeX! 

Looks like there's a merge conflict here, do you mind resolving it?

LaVibeX:
Hi @nscuro conflict was resolved 👍

nscuro (approved):

PR #776, opened 2024-03-13
feat(component): add internal / external classification badge

setchy:
### Description

Add an `internal` or `external` badge to the component view and component details model.

I'd appreciate feedback/input on the correct terminology to use for this.  I went with _component_source_ for this PR, but was also contemplating _component_scope_

### Addressed Issue

Previously, there was no way through the UI to determine if a component was `internal` or `external`

### Additional Details

Screenshot 2024-03-13 at 6 36 49 AM
Screenshot 2024-03-13 at 6 36 58 AM

### Checklist


setchy:
@nscuro - do you happen to have any feedback on this enhancement suggestion?

nscuro:
@setchy Thanks for the ping! I like the idea of having this badge. But I am also not super sure about the label name. :thinking: 

We need to be cautious to not use terms that could be confused with [provenance](https://cyclonedx.org/use-cases/#provenance), which I think would be the case for *component source*. Similar story for *component origin* for example.

> [...] but was also contemplating *component_scope*

Scope might be confused with [`component.scope`](https://cyclonedx.org/docs/1.5/json/#components_items_scope) in the spec, which to be fair, DT currently does not display.

How about *classification*? Similar to how a business would *classify* documents as confidential or not. In our case, components would be classified as either internal or external.

setchy:
Exactly the type of feedback I needed!  Really like the suggestion to use `Classification`.  Will update asap

setchy:
Updated terminology to use `Classification`

nscuro (approved):
Thanks!

PR #797, opened 2024-03-31
fix: v-for explicit key

setchy:
### Addressed Issue

Fixes startup warning

```
Module Warning (from ./node_modules/vue-loader/lib/loaders/templateLoader.js):
(Emitted value instead of an instance of Error) <router-link v-for="item in section.children">: component lists rendered with v-for should have explicit keys. See https://vuejs.org/guide/list.html#key for more info.

 @ ./src/views/administration/AdminMenu.vue?vue&type=template&id=45a3a857&scoped=true& 1:0-438 1:0-438
 @ ./src/views/administration/AdminMenu.vue
 @ ./node_modules/babel-loader/lib??ref--12-0!./node_modules/cache-loader/dist/cjs.js??ref--0-0!./node_modules/vue-loader/lib??vue-loader-options!./src/views/administration/Administration.vue?vue&type=script&lang=js&
 @ ./src/views/administration/Administration.vue?vue&type=script&lang=js&
 @ ./src/views/administration/Administration.vue
 @ ./src/router/index.js
 @ ./src/main.js
 @ multi (webpack)-dev-server/client?http://192.168.1.108:8080/sockjs-node (webpack)/hot/dev-server.js ./src/main.js
 ```

### Checklist

### DIFF
--- a/src/views/administration/AdminMenu.vue
+++ b/src/views/administration/AdminMenu.vue
@@ -24,6 +24,7 @@
       >
         <div class="list-group" id="list-tab" role="tablist">
           <router-link
+            :ref="item.id"
             :key="item.id"
             v-for="item in section.children"
             class="list-group-item list-group-item-action"
--- a/src/views/administration/AdminMenu.vue
+++ b/src/views/administration/AdminMenu.vue
@@ -24,7 +24,7 @@
       >
         <div class="list-group" id="list-tab" role="tablist">
           <router-link
-            :ref="item.id"
+            :key="item.id"
             v-for="item in section.children"
             class="list-group-item list-group-item-action"
             data-toggle="list"

### CODE REVIEW

File: src/views/administration/AdminMenu.vue [hunk 1]
```diff
@@ -24,7 +24,7 @@
       >
         <div class="list-group" id="list-tab" role="tablist">
           <router-link
-            :ref="item.id"
+            :key="item.id"
```
nscuro, line 27: My experience with Vue is limited, but based on the docs it doesn't seem like [`ref`](https://v2.vuejs.org/v2/api/#ref) and [`key`](https://v2.vuejs.org/v2/api/#key) are the same thing, semantically.

I can't really tell if `ref` was ever needed here, but on first look it seems wrong to simply replace it with `key`, instead of simply adding `key`. Any thoughts?

nscuro (approved):

PR #801, opened 2024-04-03
Feature: add support for serving from custom path

Squixx:
### Description
I want to run dependency track frontend on a path instead of `/`, but doing this currently would require to do a full rebuild of the frontend. This PR could be a possible solution direction. 

I've seen the reference to:

```
 // Relative paths cannot be supported. Research by @nscur0 - https://owasp.slack.com/archives/CTC03GX9S/p1608400149085400
  publicPath: '/',
``` 

However I can not open this link, so am oblivious to any previous research

I'll update the docs once it's clear this is the way forward

### Addressed Issue

https://github.com/DependencyTrack/dependency-track/discussions/1514

### Additional Details

setting the `BASE_HREF` env variable allows you to set a path 

### Checklist


fnxpt:
similar to #23

Squixx:
> similar to #23

Does that work with oidc? This pr also has working oidc, but I'd be happy if either could move forward

nscuro:
Sorry for the long wait @Squixx, I'll have a look at this next and have assigned it to v4.12.

docker/Dockerfile.alpine
```diff
@@ -9,8 +9,8 @@ ENV TZ=Etc/UTC \
     LANG=C.UTF-8 \
     # Set default settings that may get overridden to empty values by
     # the entrypoint script, if not explicitly provided by the user
-    OIDC_SCOPE="openid profile email"
-
+    OIDC_SCOPE="openid profile email" \
+    BASE_HREF="/"
```
nscuro, line 13: Consider renaming this to `BASE_URL` or `BASE_PATH` so it's more in line with how the other options are worded (i.e. `API_BASE_URL`).

docker/Dockerfile.alpine
```diff
@@ -29,10 +29,13 @@ RUN chown -R 101:0 ${APP_DIR} \
 # See https://github.com/nginxinc/docker-nginx-unprivileged/blob/main/stable/alpine/Dockerfile#L139
 USER 101
 
+RUN mkdir /etc/nginx/templates
```
nscuro, line 32: Ha, TIL about [config templates](https://devopsian.net/p/nginx-config-template-with-environment-vars/), nice!

nscuro (commented):
Again sorry for the delay! Overall this looks good to me. 

Just a minor nitpick about naming of the env var.

Also, it would be good to add a few words about how to configure this in the docs: https://github.com/DependencyTrack/dependency-track/blob/master/docs/_docs/getting-started/configuration.md#frontend

docker/Dockerfile.alpine
Squixx: BASE_PATH probably makes most sense, and would trigger anyone to put in the full URL. Updated

Squixx:
> Again sorry for the delay! Overall this looks good to me.
> 
> Just a minor nitpick about naming of the env var.
> 
> Also, it would be good to add a few words about how to configure this in the docs: https://github.com/DependencyTrack/dependency-track/blob/master/docs/_docs/getting-started/configuration.md#frontend

added docs in https://github.com/DependencyTrack/dependency-track/pull/3899

nscuro (approved):

Issue #851, opened 2024-05-08
Review / Improve French Translation


nscuro:
### Current Behavior

The current French translation is machine-generated, and most definitely will contain mistakes.

### Proposed Behavior

Have a native speaker review and improve the translation.

The translation file is located here: https://github.com/DependencyTrack/frontend/blob/master/src/i18n/locales/fr.json

Instructions for working on translations is here: https://github.com/DependencyTrack/frontend?tab=readme-ov-file#adding-or-improving-translations

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [X] I have checked the [existing issues](https://github.com/DependencyTrack/frontend/issues) for whether this enhancement was already requested

molusk:
Hi,
I will work on this one but I may miss some translations. I have already some that seem odd but I don't know the right terminology.
Anyway, I think I can improve the current translation :)

Thank you for integrating this feature, it will help some of my colleagues to better understand this great tool !

msymons:
@molusk, do not worry about being incomplete when it comes to correcting the translation of keys.  Every little bit helps.

When it come to keys where the existing translation is incorrect (or just odd) then simply identifying which those are is actually half the battle.  I would suggest just trying your best and then we can see about getting a PR review from another French speaker.

molusk:
Thank’s @msymons, that’s my plan, I think I may be able to make that PR today, I already reviewed about 750 keys.

Some of them were source of debate in my team without getting any solution in french (the OWASP panel in particular is very tricky to understand and translate). In some cases, we may end up keeping the english keys rather than a poorly chosen french ones.

I also noticed some keys were not translated in english, I will check whether someone else did it before or I should also issue a PR.

molusk:
I made the PR, any french speaker is welcome to review it but I think my translation is already way better than the automatic one :wink: 
There are still few things I am not happy with, all comments are welcome.

PR #878, opened 2024-05-24
Updates french translation

molusk:
### Description

First proposal for a human made french translation for Dependency Track frontend.
It may not be perfect but at least I corrected ambiguous, false and counter-meaning auto-translations.

Maybe another native french speaker could review my proposal but that first approximation is way better than the auto-translation anyway !

### Addressed Issue

#851 
I don't think we should close the issue, yet.

### Checklist


molusk:
Ok, I can see the issue with DCO, I thought I corrected the commit but it seems I did not...

nscuro (approved):
Thanks so much! 🙌🇫🇷

PR #1024, opened 2024-10-01
configuration issue updated

itzanway:
### Checklist


src/main.js
```diff
@@ -69,6 +69,13 @@ axios
     createVueApp();
   });
 
+ const faviconPath = process.env.DT_FAVICON_PATH || '/default/favicon.ico';
```
nscuro, line 72: I fear this won't work.

[Certain environment variables defined at build-time can be statically embedded](https://cli.vuejs.org/guide/mode-and-env.html#environment-variables), but `DT_FAVICON_PATH` doesn't fall under any of them. Also, even if it was, it would mean that users would have to re-compile the project if they want to change the favicon, which is not what we want.

Is there another way to achieve this?

nscuro (changes requested):

PR #1055, opened 2024-10-10
Fix: navigation in new tab and/or window

Squixx:
### Description

Currently, when running dependency track on any path other than /, whenever you open any link in 'new tab' or new window, you will be directed to the application on your base domain. Making opening up several issues etc a bit annoying.

By making all internal links relative to the root of the application (index.html) browsers will actually resolve these links relative to the current path e.g. domain.com/depenencytrack/<link> and it will actually work
 
### Addressed Issue

Havent actually created an issue for this yet

### Checklist


nscuro:
Thanks! I think a safer approach would be to use the Vue router instead of hardcoding URL paths, since it should work in all cases.

You'll want to use `router.push`, like so:

https://github.com/DependencyTrack/frontend/blob/58f52316c833c84060afbc8e8a511d3103339bfa/src/views/portfolio/projects/Component.vue#L162

It's also possible to reference routes by name, e.g.:

https://github.com/DependencyTrack/frontend/blob/58f52316c833c84060afbc8e8a511d3103339bfa/src/shared/common.js#L45

Relevant router docs are here: https://router.vuejs.org/guide/essentials/navigation.html#Navigate-to-a-different-location

For `formatter` functions specifically, as commonly used in tables, you might need to utilize this hack to gain access to the router: 

https://github.com/DependencyTrack/frontend/blob/58f52316c833c84060afbc8e8a511d3103339bfa/src/views/portfolio/projects/ProjectList.vue#L254-L256

The routes are defined here: https://github.com/DependencyTrack/frontend/blob/58f52316c833c84060afbc8e8a511d3103339bfa/src/router/index.js#L113

Squixx:
I tried this, however that seems to break internal navigation, as when i click a link within the app it'll navigate to /basepath/basepath/path

while the links correctly translate to /basepath/path in the dom. So seems the router is doing something internally. any ideas?

(see [https://github.com/Squixx/frontend/tree/feat/use-router](this branch))

PR #1066, opened 2024-10-19
fix: breadcrumbs navigation broken for non-english language

### LINKED ISSUE #880: Breadcrumbs are not working if you switch language i.e. to German
### Current Behavior

Breadcrumbs are not working if you switch language i.e. to German
Name is translate and will not work.
Name, label and path should be dedicated properties.

### Steps to Reproduce

1. Switch to German
2. Got to projects and open a project
3. or stay at projects and click on "Heim"
4. Navigate using breadcrumbs on top of the page

### Expected Behavior

navigation should go to Dashboard or to project list, if you click on "Projekte".

By the way "Heim" is an incorrect translation for the home-button. It's called "Start" in apps using a Germen lacalization.

4.11.1

### Browser

Microsoft Edge

### Browser Version

_No response_

### Operating System

Windows

### Checklist

- [X] I have checked the [existing issues](https://github.com/DependencyTrack/frontend/issues) for whether this defect was already reported

IdrisGit:
### Description

coreui/vue breadcrumbs take the name prop as the name of the component to redirect to and label to display the name in UI, since only the name prop was provided it would display that name and use that as the component name.

Since the component names are constant values in English in the Router config, it was causing a not-found issue for languages other than English. The router was not able to find components for those labels, as they were non-existent, causing the router to break.

In this change, I have added the sectionName as a constant value to be used in the name prop and i18n translation for the label prop in the breadcrumbs.

### Addressed Issue

Fixes #880 

### Additional Details

I have used this approach as it is a good practice to keep the component and route names a constant value and only use i18n for the display label, as translations can change over time and we don't want to deal with constant updates, using constant values helps keep things organised.

### Checklist

- [ ] ~~This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly~~


nscuro (approved):
Thanks!

Issue #1146, opened 2025-01-21
Tag Dropdown: Occasional Contrast Misbehaviour


msymons:
### Current Behavior

The Tag field on the "View Details" dialog for a project offers a drop-down that displays hints for adding tags.  Normally the drop-down displays with a good contrast.  But sometimes (for reasons unknown) things do not work quite right...

![Image](https://github.com/user-attachments/assets/264c6b11-1e5e-4fc6-b818-f84e215ea9aa)

### Steps to Reproduce

- Make a note of an existing tag that is used in your system.  Ideally, choose one where multiple tags start
- On project screen, click "View Details"
- Type the first letter or two of your tag into the Tags field.
- A drop-down will now display, showing you tags that match what you just typed
- The visual contrast in the drop-down should be good.  ie,  the text should be easy to read.

### Expected Behavior

The visual contrast in the "Tags" drop-down should always be good.  ie,  not sometimes good and sometimes bad.

### Dependency-Track Frontend Version

4.12.2

### Browser

Mozilla Firefox

### Browser Version

134.0.1

### Operating System

Windows

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/frontend/issues) for whether this defect was already reported

jayolee:
Please assign this to me.
It happens when the [autocomplete dropdown library](https://github.com/JohMun/vue-tags-input)(`@johmun/vue-tags-input`)'s style overrides the app's custom style.
I am not sure what triggers this though. I can force the custom style to override the dropdown library's.

The top style comes from the dropdown library, and the bottom one is the custom style from the app.
![Image](https://github.com/user-attachments/assets/2cc48d00-afad-4f7a-8e19-2870e086bb6a)

PR #1213, opened 2025-04-01
Fix tag autocomplete dropdown library style overriding issue

jayolee:
### Description

Updated tag dropdown autocomplete's style selector to prevent the input library's style overriding the custom style from the app which made it hard to read due to low contrast

### Addressed Issue

Closes: #1146 

### Additional Details

The issue is caused by the library's default style occasionally overriding the custom style.
`.ti-autocomplete` is always a `div` tag [(source code)](https://github.com/JohMun/vue-tags-input/blob/294026aeed583154df3aeaf9f8007136404c132e/vue-tags-input/vue-tags-input.vue#L150-L154) so added the tag name to the style selector to override the default style.

Screenshot 2025-03-31 at 10 41 27 PM

### Checklist

### DIFF
--- a/src/assets/scss/vendors/vue-tags-input/_vue-tags-input.scss
+++ b/src/assets/scss/vendors/vue-tags-input/_vue-tags-input.scss
@@ -25,7 +25,7 @@
 }
 
 /* some stylings for the autocomplete layer */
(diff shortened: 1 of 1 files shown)


nscuro (approved):
Thanks!

PR #1275, opened 2025-06-11
feat(policy): add tag exclusion functionality in PolicyList

arjavdongaonkar:
### Description

Introduce a boolean flag invertTagMatch to allow policies to apply to all projects, excluding those with the selected tags.

### Addressed Issue

Policies apply to projects with selected tags. There's no way to apply a policy to all projects except those with specific tags, making exclusion use cases cumbersome.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#pull-requests)
- [X] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly


nscuro (changes requested):
Thanks @arjavdongaonkar! Please add i18n for the new fields as per https://github.com/DependencyTrack/frontend#adding-or-improving-translations.

Issue #1295, opened 2025-07-16
API Key overview breaks on smaller screen sizes


Darianisak:
### Current Behavior

In the case of smaller screen sizes, e.g., below 1600px wide, Dependency Track's API Key overview (`"Administration -> Access Management -> Teams"`), will 'escape' from it's parent div, resulting in a visually broken page, occassionally with overlapping buttons.

The following screenshot, which features _now destroyed, local development tokens_, showcases this defect.

<img width="1033" height="1287" alt="Image" src="https://github.com/user-attachments/assets/daec1817-23f1-4cfc-b7db-20c5273258ce" />

While most screens are *likely* to be large enough to accommodate this, it is still a bit clunky when, for example, using non-fullscreened tabs.

### Steps to Reproduce

1. Start a local development environment and go through the "first time" admin login process.
2. Navigate: "Administration -> Access Management -> Teams"
3. Click the "Administrators" team.
4. Create an API key for the "Administrators" team.
5. Gradually reduce the width of the screen. My testing has shown the element overflow occurs below a width of 1600px.

If it's useful, this is the `docker compose` spec I have used to run up this reproduction instance:

```
# Docker Compose spec for Dependency Track (SCA) local development.
#
# Note: This compose spec is based *heavily* on the default upstream spec for
# *deploying* with Docker. For further details, please see:
#   https://docs.dependencytrack.org/getting-started/deploy-docker/
#
services:
  apiserver:
    container_name: sca_api
    hostname: sca.api
    profiles:
      - dev
    depends_on:
      database:
        condition: service_healthy
    image: dependencytrack/apiserver:latest

    # `services.apiserver.environment`
    #
    # Please refer to the DependencyTrack docs for further configuration:
    #   https://docs.dependencytrack.org/getting-started/configuration/
    #
    environment:
      ALPINE_DATABASE_MODE: "external"
      ALPINE_DATABASE_URL: "jdbc:postgresql://sca.database:5432/dtrack"
      ALPINE_DATABASE_DRIVER: "org.postgresql.Driver"
      # Note: The default admin credentials can be found here:
      #   https://docs.dependencytrack.org/getting-started/initial-startup/
      #
      # The system requires these to be changed after initialization, so we use
      # volumeMounts to preserve the changed password as it saves on time.
      #
      # You can 'reset' the password to default by deleting the volumes:
      #   docker compose --profile dev down --volumes \
      #   && docker compose --profile dev up
      #
[...]

PR #1296, opened 2025-07-16
Improve usability of the API key overview

Darianisak:
### Description

This MR amends the ApiKeyListGroupItem component to be more friendly to smaller screen sizes.

Prior to this MR, the API Keys within a Teams detail view would 'break out' of their parent div when viewing on smaller screen sizes ( < 1700px ), resulting in a clunky UX.

These overflows, while mainly a visual defect, could cause multiple buttons to share the same location, hindering accessibility.

These changes help ensure that this content is always rendered within the parent div, regardless of screen size.

### Addressed Issue

#1295

### Additional Details

N/A

### Checklist

### DIFF
--- a/src/views/administration/accessmanagement/ApiKeyListGroupItem.vue
+++ b/src/views/administration/accessmanagement/ApiKeyListGroupItem.vue
@@ -1,8 +1,8 @@
 <template>
   <b-list-group-item class="flex-column align-items-start">
     <div class="d-flex w-100 justify-content-between">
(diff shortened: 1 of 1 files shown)


hidde-jan:
I ran into the same issue. This would better be solved by changing the col-sm-6 class of the column parent element, giving more room to the elements within it.

The reason for col-sm-6 seems to be that this allows the "Delete team" button to be floating in the right top corner of the column element.

PR #1259, opened 2025-05-27
feat: add AttributedOn Policy

arjavdongaonkar:
feat(i18n): add tooltip for attributed on in English locale and update PolicyCondition.vue

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#pull-requests)
- [x] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly

PR #1305, opened 2025-08-12
feat: add patch version handling in policy conditions and update tran…

arjavdongaonkar:
…slations

### Description

Added a **Patched Version Policy** that checks if vulnerabilities in a component have a patched version available, and compares this against policy conditions.
This allows policies to trigger when:

* A patched version **is** present (or)
* A patched version **is not** present
  based on the configured condition (`IS` or `IS_NOT`) and expected value (`true` / `false`).

### Addressed Issue
Enables policy checks for vulnerabilities based on the presence or absence of patched versions.

### Additional Details
* The evaluator looks at all vulnerabilities for a component.
* It treats any non-empty `patchedVersions` field as "patched version available".
* Invalid policy values for this subject (not `true` or `false`) are ignored with a warning in logs.
* Adds flexibility to enforce upgrade-related security policies.

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#pull-requests)
- [X] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly


arjavdongaonkar:
Same as #1259, the backend counterpart (DependencyTrack/dependency-track#5202) no longer applies on v5. Closing.

PR #1505, opened 2026-05-01
Add project filter toggles to component search

hoobio:
### Description

Adds two project-scope toggles to the Component Search page (`/components`):

- "Show inactive projects" (default on; off filters to active)
- "Only show latest project versions" (default off; on filters to latest)

Each maps to a query parameter on `GET /v1/component/identity` (`excludeInactiveProjects`, `onlyLatestProjectVersions`).

### Addressed Issue

Refs DependencyTrack/dependency-track#4570

### Additional Details

Backend support: DependencyTrack/dependency-track#6085. The Impact Analysis doc update lives in the backend PR alongside the API param description.

The "Show inactive projects" toggle reuses the existing `show_inactive_projects` i18n key and matches the convention used on `ProjectList`, `PolicyViolationAudit`, `VulnerabilityAuditByOccurrence`, `VulnerabilityAuditGroupedByVulnerability`, and `AffectedProjects` (default on, off to filter). The "Only show latest project versions" toggle uses positive framing (default off, on to filter) since negation ("Show non-latest...") read awkwardly. The new `only_show_latest_project_versions` key is seeded with `null` in non-English locales per the convention used in [cfd0b1b](https://github.com/DependencyTrack/frontend/commit/cfd0b1b72af1ad2f6df489752e2f428eb7275fb0) ("Add CVSSv4 support"), pending machine translation via i18n Ally.

The page saves toggle state to `localStorage` so the chosen scope persists across sessions.

### Preview

image

### Checklist

- [x] This PR introduces new or alters existing behavior, and I have updated the documentation accordingly


hoobio:
Open question on placement: the two toggles currently sit on a new row directly below the search inputs, which adds vertical space to the page (the original issue notes there's a desire to keep the page compact and even hinted at dropping the portfolio widgets at the top to make room).

A few options I considered:

1. **New row below search inputs** (current). Simple, readable, but adds a row.
2. **Inline with the search row** (after the Search button or after the column-toggle/refresh icons). Compact, but cramps the row on narrower viewports.
3. **Inline with the bootstrap-table toolbar** (right of the column-show / refresh icons that the table already exposes). Mirrors how `ProjectList.vue` puts its `show_inactive_projects` switch alongside the create button.

I went with option 1 to avoid touching the existing search row layout. Happy to move them if there's a preferred home.

src/views/portfolio/components/ComponentSearch.vue
```diff
@@ -51,6 +51,28 @@
...
+          }}</span>
+          <c-switch
+            id="showNonLatestProjectVersions"
+            color="primary"
+            v-model="showNonLatestProjectVersions"
+            label
+            v-bind="labelIcon"
+          /><span class="text-muted">{{
+            $t('message.show_non_latest_project_versions')
+          }}</span>
```
nscuro, line 73: I think this should be "Only show latest project versions" instead. "Show non-latest project versions" reads a bit odd.

src/views/portfolio/components/ComponentSearch.vue
```diff
@@ -51,6 +51,28 @@
           }}</b-button>
         </b-col>
       </b-row>
+      <b-row class="mt-2">
```
nscuro, line 54: RE: Placement.

I don't believe there is a clean way to include the new toggles in the existing filter row without full refactor. And to be honest, I wouldn't want to spend much effort on it given we already refactored it for v5: 

Image

That is to say, the current implementation is fine, given the constraints.

nscuro (changes requested):

src/views/portfolio/components/ComponentSearch.vue
hoobio: Agreed, relabelled in [026c421](https://github.com/DependencyTrack/frontend/pull/1505/commits/026c421). Toggle is now positive-framing (default off, on filters to latest), so the active and latest toggles use slightly different framing, but each reads naturally for its own semantic.

src/views/portfolio/components/ComponentSearch.vue
hoobio: Sounds good, leaving placement as-is.

hoobio:
Toggle relabel + plural rename addressed in [026c421](https://github.com/DependencyTrack/frontend/pull/1505/commits/026c421). Backend pluralization in DependencyTrack/dependency-track#6085. Ready for another look @nscuro.

src/i18n/locales/de.json
```diff
@@ -641,6 +641,7 @@
     "oidc_redirect_failed": "Beim Umleiten zum OpenID Connect-Identitätsanbieter ist ein Fehler aufgetreten",
     "only_direct_tooltip": "Nur direkte Abhängigkeiten anzeigen, transitive Abhängigkeiten ausblenden",
     "only_outdated_tooltip": "Nur Abhängigkeiten anzeigen, für die neuere stabile Versionen verfügbar sind",
+    "only_show_latest_project_versions": null,
```
hoobio, line 644: @nscuro confirming this is the right approach: seeded the new key as `null` in all 13 non-English locales matching the pattern you used in [cfd0b1b](https://github.com/DependencyTrack/frontend/commit/cfd0b1b72af1ad2f6df489752e2f428eb7275fb0) ("Add CVSSv4 support"), so the i18n CI passes without me guessing translations. Happy to swap in machine translations via i18n Ally if you'd prefer.

nscuro (approved):

PR #1553, opened 2026-05-29
Add OWASP analysis fields to FindingAudit component

fahedouch:
### Description

Same change as DependencyTrack/hyades-frontend#465, moved here since v5 dev is on this repo now. Adds the OWASP RR vector/score and the analysis source to the finding audit
  panel (read-only), goes together with the backend PR DependencyTrack/dependency-track#6210.

### Addressed Issue

N/A

### Checklist


src/views/portfolio/projects/FindingAudit.vue
```diff
@@ -126,6 +126,34 @@
...
+        label-for="input-owasp-score"
+      >
+        <b-form-input
+          id="input-owasp-score"
+          :value="owaspScore"
+          class="form-control disabled"
+          readonly
+          trim
+        />
+      </b-form-group>
```
nscuro: Currently all other ratings only show vectors. Showing the score would for OWASP RR only would introduce an inconsistency. Is this strictly required for your use case?

src/views/portfolio/projects/FindingAudit.vue
```diff
@@ -126,6 +126,48 @@
...
+          id="input-owasp-vector"
+          :value="owaspVector"
+          class="form-control disabled"
+          readonly
+          trim
+        />
+      </b-form-group>
+      <b-form-group
+        v-if="owaspScore"
+        id="fieldset-owasp-score"
```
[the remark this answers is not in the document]
nscuro: If the score is to be retained, then this should likely be addressed.

src/views/portfolio/projects/FindingAudit.vue
fahedouch: @nscuro Agreed, it's not strictly needed for the moment, the score is derivable from the vector anyway. Dropped the score field and kept only the vector to stay consistent with the other ratings. I will make a follow-up PR is score will be requested by vens users

nscuro (approved):

PR #1561, opened 2026-06-03
Drop dev dependency @cyclonedx/cyclonedx-npm

ColorfulRhino:
### Description

~~This updates the transitive dependency "libxmljs2" which before did no longer compile from source with Node >=22, breaking reproducable builds from source (e.g. nixpkgs).~~

This SBOM generator was superseded by "@cyclonedx/webpack-plugin", but was
never removed.

This also removes the transitive dependency "libxmljs2" which before did no
longer compile from source with Node >=22, breaking reproducable builds
from source (e.g. nixpkgs).

See https://github.com/NixOS/nixpkgs/issues/515284#issuecomment-4604348065

### Addressed Issue

n/a

### Additional Details

n/a

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -61,7 +61,6 @@
       "devDependencies": {
         "@babel/core": "7.24.7",
         "@babel/eslint-parser": "7.24.1",
-        "@cyclonedx/cyclonedx-npm": "1.16.2",
         "@cyclonedx/webpack-plugin": "2.0.2",
         "@vue/babel-helper-vue-jsx-merge-props": "1.4.0",
         "@vue/babel-preset-jsx": "1.4.0",
@@ -158,7 +157,6 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.24.7.tgz",
       "integrity": "sha512-nykK+LEK86ahTkX/3TgauT0ikKoNCfKHEaZYTUVupJdTLzGNvrblu4u6fa7DhZONAltdf8e662t/abY8idrd/g==",
       "dev": true,
-      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.24.7",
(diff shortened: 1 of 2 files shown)


nscuro:
Double-checking this, it looks like the dependency on `@cyclonedx/cyclonedx-npm` is dead code, since we actually use `@cyclonedx/webpack-plugin` to emit SBOMs. Can you try removing it entirely?

ColorfulRhino:
> Double-checking this, it looks like the dependency on `@cyclonedx/cyclonedx-npm` is dead code, since we actually use `@cyclonedx/webpack-plugin` to emit SBOMs. Can you try removing it entirely?

Done. It seems to build just fine without it. But not sure where/how it was used. I am not familiar with the codebase.

Maybe there are other dependencies that are no longer needed or at least could be updated? This would reduce the number of vulnerabilities and deprecation warnings from NPM. I am happy to assist.

ColorfulRhino:
Will need to update the PR title and commit message tomorrow if the removal is the solution.

nscuro:
Builds and runs fine with the dependency removed.

> But not sure where/how it was used

It's an alternative SBOM generator for NPM. It was used before the Webpack plugin was adopted and apparently never removed. It is intended to be invoked directly, not used as a library. That's why nothing breaks when removed.

> Maybe there are other dependencies that are no longer needed or at least could be updated?

I already did a sweep the other day which addressed various vulns: https://github.com/DependencyTrack/frontend/pull/1560

ColorfulRhino:
>  It's an alternative SBOM generator for NPM. It was used before the Webpack plugin was adopted and apparently never removed. It is intended to be invoked directly, not used as a library. That's why nothing breaks when removed.

Sounds good, thanks for the explanation!

> > Maybe there are other dependencies that are no longer needed or at least could be updated?
> I already did a sweep the other day which addressed various vulns: https://github.com/DependencyTrack/frontend/pull/1560

I will try to address some more vulnerabilities and see what I can do via `npm update` if that's alright.

The good thing is, with this PR alone we went from

```
142 vulnerabilities (15 low, 73 moderate, 42 high, 12 critical)
```

to

```
135 vulnerabilities (15 low, 72 moderate, 39 high, 9 critical)
```

I know you said the removed dependency was intended to be invoked directly, but still.

nscuro (approved):

PR #1575, opened 2026-06-05
Backport 4.14.x: Update to Node 22

ColorfulRhino:
### Description

Update NodeJS to Node 22 and drop the dev dependency `@cyclonedx/cyclonedx-npm`.

### Addressed Issue

Backport #1556 and  #1561 to 4.14.x

### Additional Details

n/a

### Checklist

- [ ] ~~This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/docs) accordingly~~


ColorfulRhino:
Would appreciate a new release for the 4.14.x version soon if possible, since this would unblock the Dependency-Track package in nixpkg :)
Thanks!

nscuro:
Were you able to verify that this change indeed unblocks your Nix issue?

ColorfulRhino:
Yes. With this applied, the package builds without errors on nixpkgs.

Without this, this is the error after removing the currently pinned Node 20, which was pinned due to the build error before, but now that Node 20 is EOL needs to be bumped in order to not be marked insecure: https://github.com/NixOS/nixpkgs/issues/515284#issuecomment-4604348065

The package for nixpkgs now builds fine with Node 22, Node 24 and even Node 26.

As you can see, the build error stems from the transitive dependency `libxmljs2`, which is removed in this PR [as you can see in the package-lock.json](https://github.com/DependencyTrack/frontend/pull/1575/changes/d12d416d2f9f62bd6372868ea44e310fe986bb67#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519L12617). It was a module that `@cyclonedx/cyclonedx-npm` depended on.

nscuro (approved):

Issue #1418, opened 2026-01-13
Add banner option for admins


ElenaStroebele:
### Current Behavior

Currently there is no implementation for a banner in DependencyTrack.

### Proposed Behavior

Banners allow to propagate important status messages, as well as useful information. 
The administrators should be able to configure a banner in the "Configuration" section within "Administration".
One should be able to dismiss a banner.

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/frontend/issues) for whether this enhancement was already requested

ElenaStroebele:
@nscuro Would you consider including this as a v5 enhancement? It would allow us to propagate tool status messages (and more). If you’re open to it, I can port the changes to the v5 codebase.🌞

nscuro:
Yep. Just please make sure to pass any user-configurable data through DOMPurify before rendering it, like we already do for the welcome message: https://github.com/DependencyTrack/frontend/blob/e4b28fdf4eba531e11a92321ae117d52359f5dd1/src/views/pages/Login.vue#L160-L164

We don't want to repeat GHSA-7xvh-c266-cfr5.

PR #1612, opened 2026-06-17
Added Banner Implementation

ElenaStroebele:
### Description
Added an admin-configurable, site-wide banner (Administration → Configuration → Banner) that supports either a simple message with a preset color scheme or fully custom, sanitized HTML, with an optional dismiss button. The banner renders for logged-in users only (hidden on login/change-password pages) . Dismissals are stored per session, so a closed banner stays hidden until the user logs in again.

### Addressed Issue
[1418](https://github.com/DependencyTrack/frontend/issues/1418)

### Additional Details
Corresponding backend PR: [6440](https://github.com/DependencyTrack/dependency-track/pull/6440)

### Checklist


src/views/administration/configuration/BannerConfiguration.vue
```diff
@@ -0,0 +1,252 @@
...
+      this.customMode = !!config.customMode;
+      this.HTMLConfig = this.customMode ? config.html || '' : '';
+    },
+    async saveChanges() {
+      const bannerConfig = this.customMode
+        ? {
+            activateBanner: this.activateBanner,
+            makeBannerDismissable: this.makeBannerDismissable,
+            customMode: true,
+            html: DOMPurify.sanitize(this.HTMLConfig),
```
nscuro: The sanitization here doesn't do much for security. Clients can call the API directly which then bypasses DOMPurify. The real security boundary is at render-time, which you already set up. I'd suggest we remove sanitization here as to now raise wrong assumptions about the stored value being safe.

src/shared/bannerConfig.js
```diff
@@ -0,0 +1,35 @@
+export const BANNER_DISMISSED_KEY = 'banner-dismissed';
+
+export function getPublicBannerConfigUrl(api) {
+  return `${api.BASE_URL}/${api.URL_CONFIG_PROPERTY}/public/banner/config`;
+}
+
+export function buildExampleBannerTemplate(includeClose) {
+  const close = includeClose
+    ? `<button type="button" class="banner-close" aria-label="Close banner" style="position:absolute;right:1rem;top:50%;transform:translateY(-50%);background:transparent;border:none;font-size:1.25rem;cursor:pointer;padding:0;line-height:1;color:#fff"><i class="fa fa-times"></i></button>`
+    : '';
```
nscuro: I think it would be better to add the button code at render-time. Currently everyone who wants to provide custom banner HTML is responsible for making sure the button is included and properly styled.

Treating the HTML as "content only" and adding the button at the render site would be more straightforward for users and easier to manage. Something like this in `Banner.vue`:

```vue
<div v-else-if="visible && customMode" ref="banner" class="banner-custom">                                                                                                                                 
  <div v-html="sanitizedHTML"></div>                                                                                                                                                                       
  <button v-if="dismissable" class="banner-close" type="button" :aria-label="$t('admin.banner_dismiss')" @click="close">                                                                                                                                         
    <i class="fa fa-times"></i>                                                                                                                                                                            
  </button>                                                                                                                                                                                                
</div>    
```

So more-or-less what the non-HTML config does already.

src/shared/bannerConfig.js
```diff
@@ -0,0 +1,35 @@
+export const BANNER_DISMISSED_KEY = 'banner-dismissed';
+
+export function getPublicBannerConfigUrl(api) {
+  return `${api.BASE_URL}/${api.URL_CONFIG_PROPERTY}/public/banner/config`;
```
nscuro: It seems the banner is only displayed for authenticated users. In that case, the config property should *not* be public. This requires a small API server change to mark the new config as non-public.

src/views/administration/configuration/BannerConfiguration.vue
```diff
@@ -0,0 +1,252 @@
...
+        <b-card-body>
+          <p>{{ $t('admin.banner_explanation') }}</p>
+          <b-form-group
+            label="Banner"
+            label-size="lg"
+            label-class="font-weight-bold pt-0 mb-2"
+          >
+            <div id="activateBanner">
+              <c-switch
+                id="activateBanner"
```
nscuro: Nit: `div` and `c-switch` have the same duplicate `id`.

nscuro (changes requested):
Few small adjustments I'd like to see, but overall looks good.

src/shared/bannerConfig.js
ElenaStroebele: Good idea! I’ve also adjusted that the button color matches the corresponding text color. 🌞

ElenaStroebele:
@nscuro thanks for your feedback! As I am new to frontend development, this is super helpful 😊

src/shared/bannerConfig.js
nscuro: Made a mistake here, the non-public endpoint requires `SYSTEM_CONFIGURATION` permission so it won't work as I thought. I'll address this myself after merge since I caused the confusion to begin with.

nscuro (approved):
Thanks!

I think this would deserve a [guide](https://dependencytrack.github.io/docs/next/guides/), if you'd like to add one. The docs code is at https://github.com/DependencyTrack/docs and has an extensive linter / agent guidance setup, so you can let an agent do the writing given a bit of context and ideally some screenshots. Up to you.

ElenaStroebele:
@nscuro just finished the guide and added a PR in the docs repo 🌞

PR #1741, opened 2026-08-11
Enhance policy conditions view to display properly in case policy condition has 100k characters

### LINKED ISSUE #1738: Improve the display of large policies via bundle on UI
## Description

Large bundle-managed vulnerability policy conditions can contain close to
100,000 characters on a single logical line. When displayed with line wrapping,
CodeMirror 6 may render large blank gaps while scrolling.

The complete condition is returned by the API. This is a frontend rendering
issue caused by CodeMirror's height estimation and viewport virtualization for
extremely long wrapped lines.

## Steps to reproduce

1. Import a vulnerability policy bundle containing a condition composed of
   thousands of clauses joined by `||` on one line.
2. Open the policy details page.
3. Navigate to the Condition tab.
4. Scroll through the condition.

## Current behavior

Large blank areas appear between rendered sections. Syntax highlighting may
also appear incomplete near the end.

## Expected behavior

The complete condition should be readable with syntax highlighting and without
blank rendering gaps. Viewing it must not modify the stored condition.

## Proposed enhancement

For bundle-managed conditions whose longest line exceeds a safe threshold:

- Create a display-only multiline representation by inserting line breaks at
  CEL `||` and `&&` operators outside strings and comments.
- Preserve the raw API condition unchanged.
- Render the formatted representation in read-only CodeMirror.
- Disable wrapping as a fallback if a formatted line remains excessively long.
- Provide actions to copy or download the raw condition.
- Leave UI-managed policy creation and editing unchanged.

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/main/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/frontend/issues) for whether this enhancement was already requested

mehab:
### Description

This PR addresses a gap in the view of vulnerability policies from bundles if the policy condition has 100k characters. 
I have observed large blank spaces while viewing such large policy with the current implementation. 
The condition is effectively one single logical line (thousands of project.name.startsWith(...) || chained together with no newlines). 
 For a single line that wraps into thousands of visual rows, CodeMirror's per-line height estimation becomes very inaccurate, which produces visible large blank "gap" between the rendered chunk and the remainder. There is a  red/uncolored tail seeb at the end ([project.name.st](http://project.name.st/)) is the JS highlighter/parser hitting its limit on such an enormous single line and stopping tokenization partway. These are both known CodeMirror behaviours with extremely long single lines — not a bug in the stored data.

### Addressed Issue

Closes #1738

### Checklist

- [x] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/docs) accordingly


nscuro:
I don't think we should try to format CEL without actual AST context (which would require parsing it first).

Why don't we just:

* disable line-wrapping for conditions with long lines, *or*
* switch to a simple [`<pre>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/pre) block if we detect excessively long lines?

> The condition is effectively one single logical line (thousands of `project.name.startsWith(...) ||` chained together with no newlines).

I'd recommend to change this format to `["prefixA", "prefixB"].exists(pre, project.name.startsWith(pre))`, which is cheaper to parse, cheaper to evaluate, and requires a lot fewer characters.

I assume the condition is generated, so another thing you could do is pre-format it during generation to something like:

```js
[
  "prefixA",
  "prefixB"
].exists(
  pre,
  project.name.startsWith(pre)
)
```

Which would also make it easier to diff changes in your source of truth.

mehab:
Hey @nscuro  thank you for suggesting the macro :) It will really help. I had originally ended up with a 1.6MB file with so many conditions and I had applied refactoring but only ended up with 864 KB and so many characters. Will apply it in the backend

nscuro (approved):

PR #1780, opened 2026-08-25
Fix dependency graph purl labels

### LINKED ISSUE #1779: Dependency graph node labels show percent-encoded purls (%40 for npm scopes)
### Current Behavior

In the project Dependency Graph tab, component nodes are labelled with the raw Package URL, so the percent-encoding mandated by the purl spec leaks into the UI. Scoped npm packages therefore render as %40:

```
pkg:npm/%40dnd-kit/core@6.3.1
pkg:npm/%40floating-ui/react@0.27.20
```
Every other view that renders the same component avoids this. The component detail page shows `@dnd-kit ▸ core ▸ 6.3.1`, i.e. `group ▸ name ▸ version`.

Cause is in [src/views/portfolio/projects/ProjectDependencyGraph.vue](https://github.com/DependencyTrack/frontend/blob/master/src/views/portfolio/projects/ProjectDependencyGraph.vue) — createNodeLabel() returns purlCoordinates / purl verbatim for components, and only falls back to the readable groupId/name/version composition when neither is present:

```
createNodeLabel: function (identity) {
  // Could be a project or a directDependency object.
  // Projects don't have the objectType property.
  const isProject = !identity.objectType;
  if (!isProject && identity.purlCoordinates) {
    return identity.purlCoordinates;
  } else if (!isProject && identity.purl) {
    return identity.purl;
  } else {
    // groupId + name + version
  }
}
```

Component.vue (~L213) builds the readable label instead:

```
let label = this.component.name;
if (this.component.group) label = this.component.group + ' ▸ ' + label;
if (this.component.version) label = label + ' ▸ ' + this.component.version;
```
This affects any ecosystem whose namespace contains characters the purl spec requires to be encoded — most visibly npm scopes (@ → %40), but also Maven/Go paths with encoded segments.

See image

Image

### Steps to Reproduce

1. Upload a CycloneDX BOM for an npm project that depends on at least one scoped package (e.g. @dnd-kit/core).
2. Open the project and select the Dependency Graph tab.
3. Observe the node labels for the scoped packages: pkg:npm/%40dnd-kit/core@6.3.1.
4. Click the node to open the component detail page and observe the readable label there: @dnd-kit ▸ core ▸ 6.3.1.

### Expected Behavior

Graph node labels should be human readable, consistent with the rest of the UI. Either:

decode for display — return decodeURIComponent(identity.purlCoordinates) → pkg:npm/@dnd-kit/core@6.3.1 (minimal change, keeps the purl shape people are used to in the graph), or
reuse the same group ▸ name ▸ version composition as Component.vue for full consistency.
Happy to open a PR for whichever the maintainers prefer.

5.x

### Browser

Google Chrome

### Browser Version

151.0.7922.108

### Operating System

Linux

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/main/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/frontend/issues) for whether this defect was already reported

nissessenap:
### Description

Dependency graph node labels no longer show percent-encoded purls. Component nodes are labelled with the purl decoded for
  display, so scoped npm packages render as pkg:npm/@dnd-kit/core@6.3.1 instead of pkg:npm/%40dnd-kit/core@6.3.1. Affects any  ecosystem whose namespace requires purl-spec encoding (npm scopes most visibly, also encoded Maven/Go path segments).

### Addressed Issue

Fixes #1779

### Additional Details

createNodeLabel() returned purlCoordinates / purl verbatim. Now runs the value through decodeURIComponent(), falling back to   the raw string if it throws on a malformed sequence. The graph is display-only, so decoding is safe here — nothing downstream   parses the label.

  Went with the decode option from the issue rather than reusing Component.vue's group ▸ name ▸ version composition: one-line   change, and the purl shape in the graph stays what users already recognize. Happy to switch to the ▸ form if maintainers   prefer full consistency with the component detail page.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/dependency-track/blob/main/CONTRIBUTING.md#pull-requests)


nissessenap:
The previous commit also included unit tests, but I saw that there was none existing so I reverted the PR. 
If you want some look at commit https://github.com/DependencyTrack/frontend/pull/1780/commits/ac1a7b108d5cef9d1963a8b1b237eef06d7d49f1, easy to revert again.

nissessenap:
This is how it works on my local build: 
image

nissessenap:
As discussed in the issue we will skip this

PR #1817, opened 2026-08-31
Add tag exclusion toggle to policy scope

arjavdongaonkar:
### Description

DependencyTrack/dependency-track#7176 added `invertTagMatch`, which applies a policy to every project that does not carry its tags. This exposes it in the policy scope UI.

The switch sits next to the existing `includeChildren` and `onlyLatestProjectVersion` switches, and the tag list label swaps between "Limit to tags" and "Excluded tags" so it is clear which way the tags are being applied.

### Addressed Issue

Frontend side of https://github.com/DependencyTrack/dependency-track/issues/5042

### Additional Details

This replaces #1275, which was raised against master before the v5 restructure. Same approach, with one addition: `syncVariables` now also assigns `invertTagMatch`, otherwise the switch goes stale after saving, since the other scope fields are assigned there and it was not.

Translations are the ones from #1275.

### Checklist

- [X] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/frontend/blob/main/CONTRIBUTING.md)

### DIFF
--- a/src/views/policy/PolicyList.vue
+++ b/src/views/policy/PolicyList.vue
@@ -172,7 +172,12 @@ export default {
                         </b-col>
                       </b-row>
                     </div>
-                    <b-form-group v-if="limitToVisible === true" id="tagLimitsList" :label="this.$t('admin.limit_to_tags')">
+                    <b-form-group v-if="limitToVisible === true" id="tagLimitsList" :label="invertTagMatch ? this.$t('admin.exclude_tags') : this.$t('admin.limit_to_tags')">
+                      <b-input-group-form-switch
+                        id="isInvertTagMatch"
+                        :label="$t('admin.invert_tag_match')"
+                        v-model="invertTagMatch"
+                      />
                       <div class="list-group">
                         <span v-for="tag in tags">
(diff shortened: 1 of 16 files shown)


src/views/policy/PolicyList.vue
```diff
@@ -172,7 +172,12 @@ export default {
                         </b-col>
                       </b-row>
                     </div>
-                    <b-form-group v-if="limitToVisible === true" id="tagLimitsList" :label="this.$t('admin.limit_to_tags')">
+                    <b-form-group v-if="limitToVisible === true" id="tagLimitsList" :label="invertTagMatch ? this.$t('admin.exclude_tags') : this.$t('admin.limit_to_tags')">
+                      <b-input-group-form-switch
+                        id="isInvertTagMatch"
+                        :label="$t('admin.invert_tag_match')"
+                        v-model="invertTagMatch"
+                      />
```
nscuro, line 180: The toggle currently doesn't do anything. All other toggles trigger a policy update, this one doesn't.

nscuro (changes requested):

src/views/policy/PolicyList.vue
arjavdongaonkar: Added a watcher calling updatePolicy like the other toggles have. It was only persisting if another field happened to trigger a save.

nscuro (approved):

